PR #2, opened 2026-02-14
Fix SDK file operations, secret decryption, and minor UI issues

sdc53:
## Summary

  - **fix: resolve double `_repo/` prefix and binary file write crash in SDK file ops** — `S3Backend` was
  double-prefixing workspace paths with `_repo/`, causing all SDK file operations (`files.read`, `files.write`, etc.) to
   404 during workflow execution. Also fixes binary file writes (PPTX, PNG) crashing on null bytes in PostgreSQL text
  columns.
  - **fix: decrypt secret config values in SDK `config.get()` endpoint** — The `/api/cli/config/get` endpoint returned
  raw Fernet-encrypted blobs for secret-type values instead of decrypting them, breaking workflows that read API keys
  and credentials via the SDK.
  - **fix: use non-empty placeholder value for disabled SelectItem in Email settings** — Radix UI `SelectItem` does not
  allow empty string as value.
  - **feat: proxy /docs and /redoc through Vite dev server** — Makes Swagger UI and ReDoc accessible at `localhost:3000`
   during development.
  - **chore: add `.claude/` and `.mcp.json` to `.gitignore`**

  ## Test plan

  - [x] Verified SDK `files.read()` and `files.write()` work in workflow executions (cloud mode)
  - [x] Verified binary file writes (PPTX) succeed without PostgreSQL encoding errors
  - [x] Verified `config.get()` returns decrypted plaintext for secret-type config values
  - [x] Verified Email settings page renders without SelectItem errors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -195,6 +195,10 @@ Desktop.ini
 # Linux
 *~
 
+# ==================== CLAUDE CODE ====================
+.claude/
+.mcp.json
+
 # ==================== OTHERS ====================
 
 AzuriteConfig
--- a/api/src/routers/cli.py
+++ b/api/src/routers/cli.py
@@ -369,7 +369,13 @@ async def cli_get_config(
         raw_value = cache_entry.get("value")
         config_type = cache_entry.get("type", "string")
 
-        if config_type == "json" and isinstance(raw_value, str):
+        if config_type == "secret" and raw_value:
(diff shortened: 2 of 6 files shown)


jackmusick (approved):
Code changes look good — tested locally with our feature branch, all 796 tests pass. Merging.

PR #4, opened 2026-02-18
feat: add auto_fill support for form fields

sdc53:
## Summary
- Adds an `auto_fill` JSONB field to `form_fields` that maps sibling field names to data provider metadata keys
- When a data provider field loads options, the first result's metadata is used to automatically populate sibling fields on the form
- Fully general-purpose — works with any data provider that returns metadata, no domain-specific assumptions

## Changes
- **Migration**: Adds nullable `auto_fill` JSON column to `form_fields` table
- **ORM/Contract**: `auto_fill: dict[str, str] | None` on FormField model and Pydantic schema
- **Frontend (FormRenderer.tsx)**: When a field with `auto_fill` config loads data provider options, iterates the mapping and calls `setValue()` on sibling fields with values from the first option's metadata
- **MCP tools**: `auto_fill` included in form field responses
- **E2E tests**: 3 tests covering creation, update persistence, and null default behavior

## Example
A form field configured as:
```json
{
  "name": "company_select",
  "type": "select",
  "data_provider_id": "<uuid>",
  "auto_fill": {
    "employee_count": "recommended_employee_count",
    "industry": "industry_name"
  }
}
```
When the data provider returns `{"metadata": {"recommended_employee_count": 150, "industry_name": "Manufacturing"}}`, the `employee_count` and `industry` fields are auto-populated.

## Test plan
- [x] E2E: auto_fill config stored and returned on form creation
- [x] E2E: auto_fill preserved after form update
- [x] E2E: auto_fill defaults to null for fields without it
- [ ] Manual: verify auto-fill triggers in browser when selecting a data provider option

### DIFF
--- a/api/alembic/versions/20260217_add_form_field_auto_fill.py
+++ b/api/alembic/versions/20260217_add_form_field_auto_fill.py
@@ -4,7 +4,7 @@ Stores a mapping of sibling field names to data provider metadata paths,
 enabling auto-population of form fields when a data provider returns results.
 
 Revision ID: 20260217_auto_fill
-Revises: 20260212_drop_old_app_tables
+Revises: 20260218_is_system_users
 Create Date: 2026-02-17
 """
 
@@ -12,7 +12,7 @@ import sqlalchemy as sa
 from alembic import op
 
 revision = "20260217_auto_fill"
-down_revision = "20260212_drop_old_app_tables"
+down_revision = "20260218_is_system_users"
 branch_labels = None
 depends_on = None
 
--- a/api/alembic/versions/20260217_add_form_field_auto_fill.py
+++ b/api/alembic/versions/20260217_add_form_field_auto_fill.py
@@ -1,0 +1,28 @@
+"""Add auto_fill JSONB column to form_fields table.
+
+Stores a mapping of sibling field names to data provider metadata paths,
+enabling auto-population of form fields when a data provider returns results.
+
+Revision ID: 20260217_auto_fill
+Revises: 20260212_drop_old_app_tables
+Create Date: 2026-02-17
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "20260217_auto_fill"
+down_revision = "20260212_drop_old_app_tables"
+branch_labels = None
+depends_on = None
+
+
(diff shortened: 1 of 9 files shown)


jackmusick:
Thanks for this — the backend plumbing (ORM, Pydantic, handler, MCP improvements) looks solid. A few things to address before we can merge:

**Must fix:**

1. **Migration `down_revision` is stale** — it points to `20260218_is_system_users` but new migrations have landed on main since then. The current head is `20260220_global_app_slugs`. This will fork the Alembic graph. Please rebase onto main and update the `down_revision` to the current head. (This is expected given the pace of changes — not a problem with your PR.)

2. **Migration uses `sa.JSON()` but the ORM model uses `JSONB`** — these create different Postgres column types. The migration should use:
   ```python
   from sqlalchemy.dialects.postgresql import JSONB
   sa.Column("auto_fill", JSONB(), nullable=True)
   ```

**Should fix:**

3. **Frontend auto-fill triggers on options load, not user selection** — the auto-fill logic in `FormRenderer.tsx` runs inside `loadDataProviders` and always fills from `options[0].metadata` (the first loaded option). When the user selects a different option, the sibling fields aren't updated. If the intent is "populate on first load only," that's fine but worth a comment. Otherwise, this should fire on the select field's `onChange`.

4. **No validation that `auto_fill` target field names exist in the form** — a typo in an `auto_fill` mapping key would silently fail at runtime. Consider adding a `model_validator` to `FormSchema` that checks target names against the form's field names.

sdc53:
All feedback addressed:

1. **Migration `down_revision` rebased** — updated to `20260220_global_app_slugs` (current upstream/main head)
2. **Migration type fixed** — changed `sa.JSON()` to `JSONB()` to match the ORM model
3. **Auto-fill now fires on selection change** — added an `applyAutoFill` callback that updates sibling fields with metadata from the selected option when the user changes their selection (not just on initial load)
4. **Added `model_validator` to `FormSchema`** — validates that `auto_fill` target field names exist in the form's field list, raises `ValueError` for invalid targets

jackmusick:
Merged manually with conflict resolution (migration chain fixed to follow #6).

PR #6, opened 2026-02-18
fix: wire allow_as_query_param through backend

sdc53:
## Summary
- The `allow_as_query_param` field was defined in the Pydantic `FormField` contract and the column exists in the database, but was never mapped in the ORM model or passed through the router/serializer
- This caused the form builder's "Allow as query parameter" toggle to silently drop the value on save
- Adds the ORM column mapping, wires it through `_form_schema_to_fields()` and `compute_form_schema()`, and includes an alembic migration for fresh installs

## Test plan
- [ ] Open form builder, toggle "Allow as query parameter" on a field, save the form
- [ ] Reload the form and verify the toggle state persists
- [ ] Create a new form with the toggle enabled and verify it saves correctly

### DIFF
--- a/api/alembic/versions/20260218_add_form_field_allow_as_query_param.py
+++ b/api/alembic/versions/20260218_add_form_field_allow_as_query_param.py
@@ -1,0 +1,28 @@
+"""Add allow_as_query_param boolean column to form_fields table.
+
+Enables per-field control of whether a field's value can be populated
+from URL query parameters.
+
+Revision ID: 20260218_allow_qp
+Revises: 20260218_oauth_audience
+Create Date: 2026-02-18
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "20260218_allow_qp"
+down_revision = "20260218_oauth_audience"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "form_fields",
+        sa.Column("allow_as_query_param", sa.Boolean(), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("form_fields", "allow_as_query_param")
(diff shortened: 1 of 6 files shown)


jackmusick:
Thanks — this correctly wires the field through the ORM, serialization, and deserialization paths. A few things to address:

**Must fix:**

1. **Migration `down_revision` is stale** — it points to `20260218_oauth_audience`, which already has a child in the chain. The current head on main is `20260220_unique_slugs`. Please rebase onto main and update the `down_revision`. (Expected given the pace of changes.)

2. **`FormIndexer.index_form()` was not updated** — the git sync import path at `api/src/services/file_storage/indexers/form.py` (around line 261) creates `FormFieldORM` objects from parsed YAML and maps each attribute individually. `allow_as_query_param` is missing, so forms imported via git sync will silently lose this value. Add `allow_as_query_param=form_field.allow_as_query_param` to the `FormFieldORM(...)` constructor in `index_form()`.

**Should fix:**

3. **Enable the existing E2E test assertion** — `tests/e2e/api/test_forms.py` line 570 already sends `allow_as_query_param: True` but skips the assertion with a "not yet in DB" comment. Since this PR adds the DB column, please remove that comment and assert the value is preserved:
   ```python
   assert text_field.get("allow_as_query_param") is True
   ```

sdc53:
All feedback addressed:

1. **Migration `down_revision` rebased** — updated to `20260220_global_app_slugs` (current upstream/main head)
2. **`FormIndexer.index_form()` fixed** — added `allow_as_query_param=form_field.allow_as_query_param` to the `FormFieldORM(...)` constructor so git sync preserves the value
3. **E2E test assertion enabled** — removed the skip comment and added `assert text_field.get("allow_as_query_param") is True`

PR #9, opened 2026-03-27
draft: propose source-backed integration definitions

MTG-Thomas:
## Problem

Bifrost already has source-backed models for most portable entities:

- workflows / tools / data providers from Python decorators
- forms from `.form.yaml`
- agents from `.agent.yaml`
- apps from app directories / `app.yaml`

Integrations are the main remaining portable entity type whose definition still mostly lives in manifest-shaped metadata.

At the same time, the platform already treats `.bifrost/` as generated/system-managed state in several places:

- `.bifrost/` edits are blocked in file operations
- `/api/files/manifest` regenerates manifests from DB state
- repo sync regenerates `.bifrost/*.yaml` from DB state

That makes integrations the awkward exception.

## What We Found

From a manifest inventory pass in the fork:

- 108 workflows
- 26 integrations
- 2 agents
- 1 app

All workflow / agent / app manifest entries already point at real source files.

So the migration risk is not orphaned source. The risk is metadata still living only in the manifest layer, especially for integrations:

- config schema
- OAuth provider definition
- entity picker wiring
- stable IDs and display metadata

This PR includes a repeatable inventory script and the resulting inventory note.

## Proposal

Add a source-backed integration definition file:

`integrations/<slug>/integration.yaml`

This file would carry portable integration metadata only:

- `id`
- `name`
- `entity_id`
- `entity_id_name`
- `list_entities_data_provider_id`
- `config_schema`
- `oauth_provider`

Runtime/environment-specific state would remain DB-backed:

- config values
- secrets
- tokens
- mappings
- org-specific overrides

Under this model, `.bifrost/integrations.yaml` becomes generated export, not authored source.

## Why This Fits Existing Architecture

This follows the same pattern already used by other entity types:

- executable entities: source decorators -> DB
- portable config entities: source YAML/files -> DB
- `.bifrost`: generated projection of DB state

This is narrower than a full manifest-system rewrite. It only closes the remaining gap for integrations.

## This PR

This PR is intentionally a pilot only.

Included:

- an inventory script and inventory note
- a portable integration definition contract/loader
- a validation script
- one pilot source-backed integration definition for DNSFilter
- unit coverage for the pilot shape

Not included:

- no runtime import behavior change
- no manifest import/export change
- no deletion of `.bifrost/integrations.yaml`
- no broad migration of existing integrations

## Review Questions

1. Does this target model fit upstream’s intended direction for repo-backed entities?
2. Is `integrations/<slug>/integration.yaml` the right source-backed home?
3. Should integration metadata stay YAML/file-backed, or would maintainers prefer a different source representation?
4. If accepted, is the next step an integration indexer / serializer path similar to forms and agents?

### DIFF
--- a/api/bifrost/integration_definition.py
+++ b/api/bifrost/integration_definition.py
@@ -1,0 +1,109 @@
+"""
+Source-backed integration definition helpers.
+
+This is a bridge layer for the fork's `.bifrost/` convergence work. It models a
+portable integration definition file that can live in normal repo source, while
+leaving runtime state (mappings, secrets, tokens, org-specific config) in the
+database.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import yaml
+from pydantic import BaseModel, Field
+
+
+class SourceIntegrationConfigSchema(BaseModel):
+    """Portable config schema item for a source-backed integration definition."""
+
+    key: str = Field(description="Config key name")
+    type: str = Field(description="string | int | bool | json | secret")
+    required: bool = Field(default=False, description="Whether this config must be set")
+    description: str | None = Field(default=None, description="Human-readable description")
+    options: list[str] | None = Field(default=None, description="Allowed values for string config")
+    position: int = Field(default=0, description="Display order in UI")
+
+
+class SourceOAuthProvider(BaseModel):
+    """Portable OAuth provider definition for a source-backed integration."""
+
+    provider_name: str = Field(description="Provider identifier")
+    display_name: str | None = Field(default=None, description="UI display name")
+    oauth_flow_type: str = Field(default="authorization_code", description="OAuth flow type")
+    client_id: str = Field(
+        default="__NEEDS_SETUP__",
+        description="OAuth client ID sentinel; never store a live secret-bearing client here",
+    )
+    authorization_url: str | None = Field(default=None, description="OAuth authorization endpoint")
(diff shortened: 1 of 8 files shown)


jackmusick:
Closing this — appreciate the thorough analysis but we don't think this fits upstream.

The core premise is that integrations should get source-backed definition files for consistency with workflows/forms/agents. But those entity types are source-backed for different reasons:

- **Workflows** are Python code — the decorator IS the definition
- **Forms/agents** are developer-authored YAML that gets indexed into the DB
- **Integrations** are platform configuration — set up once via UI or manifest sync, then managed in the DB. Nobody is hand-editing `integration.yaml` files day-to-day.

The manifest system already handles integration round-tripping (DB ↔ `.bifrost/integrations.yaml`) and the `_resolve_integration` upsert pattern is stable. Adding a second source-of-truth would mean maintaining two serialization paths and deciding which wins on conflict.

Specific concerns:
- `SourceIntegrationDefinition` duplicates fields already in `ManifestIntegration`
- `to_manifest_dict()` bridge is a red flag — if you need a translation layer to the existing model, the new model isn't adding value
- The audit script and planning docs are useful for fork analysis but not something to carry upstream

If there's a specific pain point with how integrations work today, happy to discuss that directly — but "consistency with other entity types" isn't sufficient motivation given the different authoring patterns.

PR #11, opened 2026-03-27
fix: include global oauth connections in org listings

MTG-Thomas:
## Summary
Include global OAuth connections in admin org listings.

## Problem
Platform-admin views of OAuth connections can miss global connections, which makes integration setup and troubleshooting harder for admin users working across org scopes.

## Change
- include global OAuth connections in the admin listing path
- add regression coverage for the org-listing behavior

## Why this is safe
- this only changes listing visibility for admin users
- it does not alter token storage, refresh behavior, or non-admin access control

## Validation
- targeted unit coverage added for the router behavior

### DIFF
--- a/api/src/routers/oauth_connections.py
+++ b/api/src/routers/oauth_connections.py
@@ -142,7 +142,14 @@ class OAuthConnectionRepository:
         """List all OAuth connections, optionally filtered by org."""
         query = select(OAuthProvider)
         if org_id:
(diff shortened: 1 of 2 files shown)


jackmusick:
Closing — this endpoint (GET /api/oauth/connections) is unused admin dead code. We'll remove it entirely instead of fixing it.

PR #12, opened 2026-03-27
fix: align shared path references

MTG-Thomas:
## Summary
Align shared-path references with the current repo layout.

## Problem
Several development and test references still assume an `api/shared` path that no longer matches the repo layout. That causes avoidable friction in local builds, compose-based workflows, and type-checking configuration.

## Change
- update Docker, compose, and related references to the current shared path layout
- align developer-facing documentation with the actual repo structure

## Why this is safe
- this is path/reference cleanup only
- it does not change application runtime logic
- it reduces mismatch between tooling and the checked-in repo structure

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ docker compose logs -f api
 docker compose logs -f client
 ```
 
-Changes to files in `api/src/`, `api/shared/`, and `client/src/` automatically reload.
+Changes to files in `api/src/`, `shared/`, and `client/src/` automatically reload.
 
 **VS Code Debugging:**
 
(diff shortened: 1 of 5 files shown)


jackmusick:
Closing — this PR changes all `api/shared` references to `shared/`, but there is no top-level `shared/` directory in the repo. The actual path is `api/shared/` and the current Dockerfile, docker-compose, and pyrightconfig references are already correct. Merging this would break Docker builds, dev/test volume mounts, and type checking.

The README additions about CLI `pass` credential storage also appear unrelated and document a feature that doesn't exist.

PR #17, opened 2026-04-01
fix: retry client requests with refreshed auth

### LINKED ISSUE #16: Client retries stale Authorization header after token refresh
## Summary
After a `401`, the frontend refreshes the access token successfully, but then retries the original request with `fetch(request.clone())`. If the original request already carried an `Authorization: Bearer ...` header, the retry reuses the stale token instead of the freshly refreshed one.

## Current behavior
In `client/src/lib/api-client.ts`:
- the client detects `401`
- calls `refreshAccessToken()`
- on success, retries with `fetch(request.clone())`

That replay keeps the old `Authorization` header intact.

## Expected behavior
After a successful refresh, the retried request should rebuild its auth headers from current token state before reissuing the request.

## Why this matters
This can leave the client stuck in an avoidable auth failure path even though refresh succeeded.

## Likely fix shape
Instead of replaying the original cloned request as-is, rebuild the request headers and replace/remove the `Authorization` header using the current token before retry.

## Affected file
- `client/src/lib/api-client.ts`

MTG-Thomas:
Closes #16

## Summary
Retry client requests after token refresh using the current auth state instead of replaying the stale original `Authorization` header.

## Problem
On `401`, the client refreshes successfully but then retries with `fetch(request.clone())`. If the original request already had a bearer token, the retry can reuse the stale token instead of the freshly refreshed one.

## Change
- add a small helper that rebuilds request headers from current token state
- retry the request with a refreshed `Authorization` header after refresh succeeds

## Validation
- change is isolated to `client/src/lib/api-client.ts`
- local TypeScript validation in this checkout is blocked because `tsc` is not installed in the current client toolchain

### DIFF
--- a/client/src/lib/api-client.ts
+++ b/client/src/lib/api-client.ts
@@ -114,6 +114,27 @@ async function refreshAccessToken(): Promise<boolean> {
 }
 
 /**
+ * Retry a request after token refresh using current auth state.
+ */
+async function retryRequestWithFreshAuth(request: Request): Promise<Response> {
+	const headers = new Headers(request.headers);
+	const token = localStorage.getItem(ACCESS_TOKEN_KEY);
+
+	if (token) {
+		headers.set("Authorization", `Bearer ${token}`);
+	} else {
+		headers.delete("Authorization");
+	}
+
+	return fetch(
+		new Request(request, {
+			headers,
+			credentials: "same-origin",
+		}),
+	);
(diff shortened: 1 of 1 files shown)


jackmusick (approved):
Correct fix — rebuilding headers from current token state avoids replaying the stale bearer on retry. LGTM.

PR #18, opened 2026-04-01
fix: handle fast workflow completion races

### LINKED ISSUE #15: useWorkflowMutation can miss fast workflow completions and stay unresolved
## Summary
`useWorkflowMutation()` relies on WebSocket completion updates after subscribing to an execution. If a workflow reaches terminal state before the completion callback is effectively attached, the hook can remain unresolved even though the execution already finished.

## Current behavior
In `client/src/lib/app-code-platform/useWorkflowMutation.ts`:
- execute workflow
- subscribe to execution updates/logs
- wait for WebSocket completion path

There is no immediate post-subscribe status check to catch executions that completed very quickly.

## Expected behavior
After subscribing, the hook should immediately check execution status once and resolve/reject if the execution is already in a terminal state.

## Why this matters
Fast workflows can complete before the normal completion event is observed by the hook, leaving app-code queries/mutations stuck with no result.

## Likely fix shape
Immediately fetch execution status after subscriptions are stored. If the execution is already terminal (`Success`, `Failed`, `CompletedWithErrors`, `Timeout`, `Cancelled`), resolve or reject the pending deferred state immediately.

## Affected file
- `client/src/lib/app-code-platform/useWorkflowMutation.ts`

MTG-Thomas:
Closes #15

## Summary
Handle app-code workflow executions that reach terminal state before the WebSocket completion callback is effectively attached.

## Problem
`useWorkflowMutation()` can stay unresolved for very fast workflows because it subscribes to execution updates and then waits for the WebSocket completion path without doing an immediate terminal-state check.

## Change
- define terminal statuses explicitly
- immediately fetch execution status once after subscription setup
- resolve or reject pending state immediately if the execution already completed

## Validation
- change is isolated to `client/src/lib/app-code-platform/useWorkflowMutation.ts`
- local TypeScript validation in this checkout is blocked because `tsc` is not installed in the current client toolchain

### DIFF
--- a/client/src/lib/app-code-platform/useWorkflowMutation.ts
+++ b/client/src/lib/app-code-platform/useWorkflowMutation.ts
@@ -37,6 +37,13 @@ function createDeferred<T>(): Deferred<T> {
 }
 
 const EXECUTION_TIMEOUT_MS = 5 * 60 * 1000; // 5 minutes
+const TERMINAL_STATUSES: ExecutionStatus[] = [
+	"Success",
+	"Failed",
+	"CompletedWithErrors",
+	"Timeout",
+	"Cancelled",
+];
 
 export interface UseWorkflowMutationResult<T> {
 	execute: (params?: Record<string, unknown>) => Promise<T>;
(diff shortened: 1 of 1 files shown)


jackmusick (approved):
Good fix for the fast-completion race — immediate terminal-state check after subscription setup is the right pattern. LGTM.

PR #21, opened 2026-04-08
fix: resolve OAuth refresh {entity_id} from integration settings

MTG-Thomas:
## Summary
- resolve OAuth URL templates from `integration.default_entity_id` or `integration.entity_id`
- apply the same fallback in the OAuth connections router, CLI refresh path, and scheduler refresh job
- add regression coverage for connection refresh and CLI token refresh

## Problem
The platform stores tenant-specific integration state in `entity_id`, but the OAuth refresh path only looked at `default_entity_id` when resolving `{entity_id}` in token URLs.

That meant a configured connection like Microsoft Graph could look valid in the integration API while `POST /api/oauth/connections/{connection_name}/refresh` still sent requests to a literal URL containing `{entity_id}` and failed with tenant-resolution errors.

## Verification
- `pytest -q tests/unit/routers/test_oauth_refresh.py tests/unit/routers/test_cli_refresh_token.py`
- `python3 -m py_compile src/routers/oauth_connections.py src/routers/cli.py src/jobs/schedulers/oauth_token_refresh.py`

## Notes
This was reproduced against a client-credentials Microsoft Graph connection whose integration had `entity_id` set and `default_entity_id` unset.

### DIFF
--- a/api/src/jobs/schedulers/oauth_token_refresh.py
+++ b/api/src/jobs/schedulers/oauth_token_refresh.py
@@ -124,8 +124,10 @@ async def run_refresh_job(
                         select(Integration).where(Integration.id == provider.integration_id)
                     )
                     integration = result_int.scalar_one_or_none()
-                    if integration and integration.default_entity_id:
-                        integration_entity_id = integration.default_entity_id
+                    if integration:
+                        integration_entity_id = (
+                            integration.default_entity_id or integration.entity_id
+                        )
 
                 token_refresh_data.append({
                     "token_id": token.id,
--- a/api/src/routers/cli.py
+++ b/api/src/routers/cli.py
@@ -1112,10 +1112,10 @@ async def sdk_integrations_refresh_token(
             if mapping:
                 entity_id = mapping.entity_id
             else:
-                # Fall back to integration's default_entity_id
+                # Fall back to the integration's configured entity identifier
                 integration = await repo.get_by_id(provider.integration_id)
                 if integration:
-                    entity_id = integration.default_entity_id
+                    entity_id = integration.default_entity_id or integration.entity_id
 
         # Decrypt client secret
         client_secret = None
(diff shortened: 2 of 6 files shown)


MTG-Thomas:
@jackmusick - not sure if this falls under "to be deprecated" code, but codex swears this is an upstream situation and nothing we did via our workspace. I'm skeptical, but it's up to you.

jackmusick (commented):
Fix looks good and is consistent across all three refresh sites (scheduler, CLI SDK endpoint, and oauth_connections router) — all now use \`integration.default_entity_id or integration.entity_id\` with the correct precedence. Traced the three call sites and confirmed the fallback lands in the actual token URL substitution path in each case.

Two small notes:

1. The scheduler path in \`api/src/jobs/schedulers/oauth_token_refresh.py\` has no dedicated regression test. The CLI and \`get_url_resolution_defaults\` helper are covered nicely by the new tests, but a one-shot test asserting the scheduler picks up \`integration.entity_id\` when \`default_entity_id\` is unset would close the loop and guard against divergence if one of the three sites is touched in the future.

2. Longer term, consider extracting the \`default_entity_id or entity_id\` resolution into a tiny helper on \`Integration\` (or a module function) so the three sites can't drift. Not blocking.

These paths are on the sensitive-paths list (OAuth / credential routers), so flagging for a human eye to confirm the change is strictly placeholder-resolution and not touching secret or scope handling before approval — on my read it isn't.

jackmusick:
Superseded by 5757f3aa (`refactor: unify OAuth token refresh across scheduler, SDK, and connections router`) on main.

That refactor hoists the `org mapping → integration.default_entity_id → integration.entity_id` fallback into a single helper (`build_token_refresh_context` in `src/services/oauth_provider.py`) so the scheduler, SDK, and oauth_connections refresh sites can't drift again, and adds `test_oauth_refresh_context.py` to pin the fallback chain. Closing — the underlying fix is already in main.

PR #22, opened 2026-04-21
fix(embed): isolate embed tokens to sessionStorage so they don't pollute user sessions

sdc53:
## Problem

Embed tokens minted by `/embed/apps/{slug}` (and `/embed/forms/{form_id}`) are delivered via a URL fragment and then persisted in the browser. Today the client stores them in `localStorage` under the same key (`bifrost_access_token`) as the user's normal session token.

Because `localStorage` is shared across every tab on the same origin, opening an embed — e.g. a HaloPSA Custom Tab iframe — silently overwrites the user's regular session in every other tab. Until the user explicitly logs out and back in, every other tab sees only the embedded app's narrow access scope (the embed JWT is scoped to one app + verified params), so the user appears to lose access to the rest of the platform.

Reproduced in production after wiring up the first Halo Custom Tab embed: opening the embed in Halo immediately broke every other Bifrost tab on the same browser, locking the user to the embed's single client until logout.

## Fix

Two-store token model:

- **`localStorage["bifrost_access_token"]`** — the user's normal session token (unchanged)
- **`sessionStorage["bifrost_embed_token"]`** — per-tab embed token, scoped to the iframe / tab that received it

A new `client/src/lib/auth-token.ts` module owns the storage rules:

- `consumeEmbedTokenFromHash()` — runs once at module load in `api-client.ts`. If the URL has `#embed_token=…`, strip the fragment from the URL bar and store the token in **sessionStorage** — but only when `window.top !== window.self`, so opening an embed URL in a regular tab doesn't accidentally turn that tab into a sticky embed session. The HMAC check on the server is the real security boundary; this is purely UX hygiene.
- `getActiveToken()` — used everywhere we attach an Authorization header (REST + WebSocket). Prefers the per-tab embed token, falls back to the user token. Embed iframes therefore present their embed JWT; other tabs see no change.
- `clearEmbedToken()` — called on every explicit login (password / MFA / OAuth / passkey) so a deliberate sign-in always wins, even from a tab that previously held an embed token.
- `clearAuthTokens()` — called on logout and on hard auth failures so a stale embed token can't keep an authenticated session alive after the user has signed out.

`ensureValidToken()` short-circuits when an embed token is present — embed tokens have no refresh flow (they're ~8h, single-use per app/form), so the normal-session refresh path doesn't apply.

## Files touched

- `client/src/lib/auth-token.ts` (new)
- `client/src/lib/api-client.ts` — replace inline `ACCESS_TOKEN_KEY` with helpers; auth-header sites use `getActiveToken()`; refresh path skips embed sessions
- `client/src/contexts/AuthContext.tsx` — clear embed token on every explicit login; logout uses `clearAuthTokens()`
- `client/src/services/websocket.ts` — WebSocket auth uses `getActiveToken()` instead of reading `localStorage` directly

## Why no server changes

The server already mints two distinct kinds of token (`embed: true` claim plus per-app/per-form scope). The bug is entirely in how the client stores them. Server contracts unchanged.

## Test plan

- [x] Open a Bifrost embed inside an iframe (HaloPSA Custom Tab). Confirm it loads against its scoped resources. *Verified in production.*
- [x] In a separate tab on the same browser, open the platform UI. Confirm the user's normal session and full org access are intact. *Verified in production — the bug that motivated this fix.*
- [ ] Log out from a normal tab → confirm the embed iframe also loses access on next request (both tokens cleared).
- [ ] Open the embed URL directly in a normal tab (not iframed). Confirm it does NOT install the embed token; the URL fragment is still stripped.
- [ ] Explicit re-login from a tab that previously held an embed token wins (no shadowing by sessionStorage).

### DIFF
--- a/client/src/contexts/AuthContext.tsx
+++ b/client/src/contexts/AuthContext.tsx
@@ -15,6 +15,11 @@ import {
 	type ReactNode,
 } from "react";
 import { useNavigate, useLocation } from "react-router-dom";
+import {
+	ACCESS_TOKEN_KEY,
+	clearAuthTokens,
+	clearEmbedToken,
+} from "@/lib/auth-token";
 
 // User info extracted from JWT
 export interface AuthUser {
@@ -71,7 +76,6 @@ const AuthContext = createContext<AuthContextValue | null>(null);
 
 // Token storage keys
 // Note: Refresh token is stored in HttpOnly cookie only (more secure)
-const ACCESS_TOKEN_KEY = "bifrost_access_token";
 const USER_KEY = "bifrost_user";
 
 // JWT payload structure
@@ -280,6 +284,7 @@ export function AuthProvider({ children }: AuthProviderProps) {
 
 			// Success - store access token (refresh token is in HttpOnly cookie)
 			if (data.access_token) {
+				clearEmbedToken();
 				localStorage.setItem(ACCESS_TOKEN_KEY, data.access_token);
 
 				const payload = parseJwt(data.access_token);
@@ -326,6 +331,7 @@ export function AuthProvider({ children }: AuthProviderProps) {
 			const data = await res.json();
 
 			if (data.access_token) {
+				clearEmbedToken();
 				localStorage.setItem(ACCESS_TOKEN_KEY, data.access_token);
 
 				const payload = parseJwt(data.access_token);
@@ -370,6 +376,7 @@ export function AuthProvider({ children }: AuthProviderProps) {
 			const data = await res.json();
 
 			if (data.access_token) {
+				clearEmbedToken();
 				localStorage.setItem(ACCESS_TOKEN_KEY, data.access_token);
 
 				const payload = parseJwt(data.access_token);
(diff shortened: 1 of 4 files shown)


jackmusick (approved):
Approving — the two-store model, iframe guard in `consumeEmbedTokenFromHash`, login-wins ordering (`clearEmbedToken()` before writing the new user token), and the `isEmbedSession()` short-circuit in `ensureValidToken` all read correctly. Nice that `websocket.ts` still re-checks `isEmbedToken(token)` before putting it on the query string — good defense in depth.

For context on future PRs: our standard is evolving toward requiring frontend tests for functional modules like this one. A vitest suite on `auth-token.ts` (iframe vs top-level `consumeEmbedTokenFromHash`, `getActiveToken` preference order, login/logout clearing semantics) would lock in the regression this fix addresses — the bug only reproduces across two tabs on the same origin, so a future refactor could silently undo it. Approving this time while we update the PR guidance; I'll commit follow-up tests as/if needed.

PR #25, opened 2026-04-22
Fix missing OAuth scope on authorization-code and refresh token exchanges

MTG-Thomas:
## Summary
This fixes delegated OAuth connections against Azure / Entra v2 token endpoints by forwarding configured scopes during authorization-code and refresh-token exchanges.

Today, Bifrost sends `scope` when building authorization URLs and for client-credentials token requests, but not when exchanging an auth code or refreshing an access token. For Azure v2 endpoints, that can fail with:

`AADSTS900144: The request body must contain the following parameter: 'scope'.`

## What changed
- Added optional `scopes` support to `OAuthProviderClient.exchange_code_for_token()`
- Added optional `scopes` support to `OAuthProviderClient.refresh_access_token()`
- Passed configured provider scopes from the OAuth callback path into the auth-code token exchange
- Passed configured provider scopes from the authorization-code refresh path into the refresh-token exchange
- Added regression coverage for scope forwarding in these flows

## Why this belongs upstream
This is a platform OAuth bug, not a workspace-specific integration quirk. Upstream already supports configured scopes for:
- authorization URL generation
- client-credentials token acquisition

This patch makes the authorization-code and refresh-token paths consistent with that model.

## Behavior notes
- `scope` is only included when scopes are configured
- client-credentials behavior is unchanged
- token storage and OAuth connection schemas are unchanged

## Test coverage
- unit tests for token exchange payload construction
- unit tests for refresh payload construction
- router-level verification that configured provider scopes are forwarded in the authorization-code refresh path
- router-level verification that configured provider scopes are forwarded in the OAuth callback path

## Validation
- `/home/thomas/mtg-bifrost/bifrost/.venv/bin/pytest tests/unit/services/test_oauth_provider.py tests/unit/routers/test_oauth_refresh.py -q -s`
- `24 passed in 11.65s`

## Follow-up note
A possible follow-up refinement is to have the authorization-code refresh path prefer stored token scopes over provider scopes if stricter grant preservation is desired.


MTG-Thomas:
@jackmusick -- PR for your review.

jackmusick (changes requested):
Thanks for catching this — the OAuth scope fix itself is solid, and the test coverage is thorough. Full `./test.sh` passes on the branch (3914/0) and the Azure v2 rationale checks out.

However this PR is bundling several unrelated changes that I'd like separated before merge:

1. **`api/Dockerfile`** — the `COPY api/shared/` → `RUN mkdir -p /app/shared` change breaks upstream. `api/shared/` has real code here (models, user provisioning, etc.); removing the COPY would ship images missing `/app/shared`. Looks like a fork-local accommodation that shouldn't go upstream.

2. **`.github/workflows/ci.yml`** — this is a full rewrite of the CI pipeline (collapses unit+e2e into one job, deletes the `build-dev` job, replaces `:dev` with `:main-latest`, makes main builds wait on e2e). Some of these are reasonable tradeoffs, but they're independent decisions from the OAuth fix and your branch appears to predate my `test.sh` refactor on Apr 20 — the workflow here calls `./test.sh --coverage --ci`, but `--ci` is now a subcommand (`./test.sh ci`), so the job wouldn't run as-is. Could you rebase on current main and drop the CI changes? Happy to review a separate PR for the pipeline ideas.

3. **`docs/plans/2026-04-04-cloudflare-deployment-options.md`** — unrelated 356-line doc, please pull out of this PR.

4. **`.gitignore` + `.superpowers/` + `=2.0.0` cleanup** — these are legitimate cleanup of files I accidentally committed, thank you. Could go in a separate small hygiene PR, or I can clean them up on my side — either works.

Once narrowed to just the OAuth scope changes + tests, I'll approve and merge. There's also a merge conflict against main I'd like resolved in the rebase.

MTG-Thomas:
I rebased this into a clean replacement PR from current `main` with only the intended OAuth scope changes and tests: #26

This original PR was branched from our fork's `main`, which pulled in unrelated fork-local commits. Please review/merge #26 instead.

PR #26, opened 2026-04-23
Fix missing OAuth scope on authorization-code and refresh token exchanges

MTG-Thomas:
## Summary
This fixes delegated OAuth connections against Azure / Entra v2 token endpoints by forwarding configured scopes during authorization-code and refresh-token exchanges.

Today, Bifrost sends `scope` when building authorization URLs and for client-credentials token requests, but not when exchanging an auth code or refreshing an access token. For Azure v2 endpoints, that can fail with:

`AADSTS900144: The request body must contain the following parameter: 'scope'.`

## What changed
- Added optional `scopes` support to `OAuthProviderClient.exchange_code_for_token()`
- Added optional `scopes` support to `OAuthProviderClient.refresh_access_token()`
- Passed configured provider scopes from the OAuth callback path into the auth-code token exchange
- Passed configured provider scopes from the authorization-code refresh path into the shared refresh path
- Added regression coverage for scope forwarding in these flows

## Why this belongs upstream
This is a platform OAuth bug, not a workspace-specific integration quirk. Upstream already supports configured scopes for:
- authorization URL generation
- client-credentials token acquisition

This patch makes the authorization-code and refresh-token paths consistent with that model.

## Behavior notes
- `scope` is only included when scopes are configured
- client-credentials behavior is unchanged
- token storage and OAuth connection schemas are unchanged

## Test coverage
- unit tests for token exchange payload construction
- unit tests for refresh payload construction
- router-level verification that configured provider scopes are forwarded in the authorization-code refresh path
- router-level verification that configured provider scopes are forwarded in the OAuth callback path

## Validation
- `/home/thomas/mtg-bifrost/bifrost/.venv/bin/pytest tests/unit/services/test_oauth_provider.py tests/unit/routers/test_oauth_refresh.py -q -s`
- `24 passed in 3.43s`

## Follow-up note
A possible follow-up refinement is to have the authorization-code refresh path prefer stored token scopes over provider scopes if stricter grant preservation is desired.

### DIFF
--- a/api/src/routers/oauth_connections.py
+++ b/api/src/routers/oauth_connections.py
@@ -831,6 +831,7 @@ async def oauth_callback(
         client_id=provider.client_id,
         client_secret=client_secret,
         redirect_uri=redirect_uri,
+        scopes=" ".join(provider.scopes) if provider.scopes else None,
         audience=provider.audience,
     )
 
--- a/api/src/services/oauth_provider.py
+++ b/api/src/services/oauth_provider.py
@@ -125,6 +125,7 @@ class OAuthProviderClient:
         client_id: str,
         client_secret: str | None,
         redirect_uri: str,
+        scopes: str | None = None,
         audience: str | None = None,
     ) -> tuple[bool, dict]:
         """
@@ -159,6 +160,9 @@ class OAuthProviderClient:
         if audience:
             payload["audience"] = audience
 
+        if scopes:
+            payload["scope"] = scopes
+
         return await self._make_token_request(token_url, payload)
 
     async def refresh_access_token(
@@ -167,6 +171,7 @@ class OAuthProviderClient:
         refresh_token: str,
         client_id: str,
         client_secret: str | None,
+        scopes: str | None = None,
         audience: str | None = None,
     ) -> tuple[bool, dict]:
         """
@@ -196,6 +201,9 @@ class OAuthProviderClient:
 
         if audience:
             payload["audience"] = audience
+
+        if scopes:
+            payload["scope"] = scopes
 
         return await self._make_token_request(token_url, payload)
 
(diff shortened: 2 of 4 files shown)


MTG-Thomas:
@jackmusick We cleaned up the issues from #25 and moved the fix here.

This PR is now rebased onto current `main` and narrowed to just the intended OAuth scope changes plus tests:
- `api/src/services/oauth_provider.py`
- `api/src/routers/oauth_connections.py`
- `api/tests/unit/services/test_oauth_provider.py`
- `api/tests/unit/routers/test_oauth_refresh.py`

Validation on this branch:
- `/home/thomas/mtg-bifrost/bifrost/.venv/bin/pytest tests/unit/services/test_oauth_provider.py tests/unit/routers/test_oauth_refresh.py -q -s`
- `24 passed in 3.43s`

Thanks again for catching the branch-base issue on the original PR.

jackmusick (changes requested):
Thanks for splitting this out — this is exactly what I was hoping for. The fix is surgical, tests are thorough, CI is green (Unit + E2E both passing), and it merges cleanly. Full `./test.sh` on the branch locally is all green aside from one pre-existing flake (`test_get_module_index_sync_empty` in `test_module_cache.py` — it mocks Redis but doesn't mock the S3 fallback path, unrelated to this PR).

One small ask before I approve:

**Please drop `_ensure_fastmcp_toolresult_export()` from `api/tests/unit/routers/test_oauth_refresh.py`.** Our current `fastmcp` (3.2.4) already exports `ToolResult` from `fastmcp.tools`, so the `if not hasattr(...)` guard never fires — it's a no-op shim in upstream. If your local env needs it, that's a fork-side fastmcp pin mismatch rather than something we want carried into the shared test file. Just remove the helper, its two callsites, and the two added imports at the top of the file.

Once that's removed I'll approve and merge.

MTG-Thomas:
Requested changes made @jackmusick

jackmusick (approved):
Thanks for the quick turnaround on both the scope cleanup and dropping the shim. Verified locally: 24/24 OAuth tests pass on 4c599e93, CI is green on the correct 4-job pipeline, and the PR merges cleanly.

Approving and merging.

PR #27, opened 2026-04-23
fix(workflows): honor documented timeout range

MTG-Thomas:
## Summary
This aligns workflow timeout enforcement with the documented contract.

Today the workflow contracts and generated docs say:
- `timeout_seconds` supports `0 = no timeout`
- the maximum timeout is `86400` seconds

But the live API and editor currently reject anything above `7200`, and the API also rejects `0`.

This PR makes runtime behavior match the documented/API contract by:
- allowing `timeout_seconds` in the `0-86400` range in workflow validation
- allowing `timeout_seconds` in the `0-86400` range in workflow update handling
- updating the workflow editor input/help text to match the same range
- adding an e2e regression test covering both `86400` and `0`

## Why
These surfaces were out of sync:
- contracts/docs advertised `0-86400`
- runtime validation enforced `1-7200`
- the editor UI enforced `1-7200`

That mismatch is user-visible and caused valid-looking values from the docs to be rejected at runtime.

## Validation
- `python3 -m py_compile api/src/services/workflow_validation.py api/src/routers/workflows.py api/tests/e2e/api/test_workflows.py`
- grep sweep on touched source/test surfaces for stale `1-7200` workflow-timeout messaging

I did not run the full e2e suite locally.


jackmusick (dismissed):
Thanks — the validation/UI changes look right and the test you added is the right test to have. But it's failing for a real reason, not flake.

**`api/src/routers/workflows.py:110`** has `timeout_seconds=workflow.timeout_seconds or 1800` in the response builder. `0` is falsy in Python, so the API response coerces `0` back to `1800`. That's why the new test sees `1800 == 0` fail: the DB correctly persists 0, the validation correctly accepts 0, but the response lies on the way out. User-visible consequence: a workflow with timeout=0 in the DB renders as "1800" in the editor input, and the next save round-trips back to 1800.

Suggested fix:

```python
timeout_seconds=workflow.timeout_seconds if workflow.timeout_seconds is not None else 1800,
```

That matches the pattern already in `api/src/services/execution/service.py:103`, `:116`, `:179`. Once that lands, your existing test should pass without further changes.

Worth flagging while we're here (not blocking, separate PR is fine): line 118 has the same `or 300` pattern for `cache_ttl_seconds`. If 0 ever becomes meaningful for caching too, same bug. Up to you whether to bundle the fix or punt.

The execution engine itself handles 0 correctly already (`service.py:443` gates on `if timeout_seconds > 0`), so this is purely a response-serialization bug, not an engine bug.

jackmusick:
Hey @MTG-Thomas — pushed the one-line fix on your behalf since it's been a few days. The change at routers/workflows.py:107 swaps `workflow.timeout_seconds or 1800` for an explicit `None` check so `timeout_seconds=0` (which the docstring documents as "disables the timeout") is actually honored. Auto-merge will fire when CI passes.

jackmusick:
Reviewed — fix is correct (the `or 1800` bug where 0 was getting silently coerced is a real find), tests cover both bounds, docs and UI updated. Refreshed against main and queued for auto-merge. Thanks for catching this.

PR #37, opened 2026-04-25
Add readiness health probes

MTG-Thomas:
## Summary

Add explicit liveness, readiness, and detailed health endpoints while keeping the existing `/health` endpoint backward-compatible.

- Keep `GET /health` as a shallow liveness response
- Add `GET /health/live` for process/API liveness
- Add `GET /health/ready` for readiness across core serving dependencies
- Expand `GET /health/detailed` with database, Redis, RabbitMQ, and configured S3 component checks
- Point Docker Compose API healthchecks at `/health/ready` with a start period for slow dev/test startup

## Why

`/health` currently answers even when dependencies needed to serve normal traffic may be unavailable. This separates liveness from readiness so Docker Compose and future orchestrators can restart or gate traffic on the right signal without breaking existing callers of `/health`.

The dependency checks use bounded async timeouts and sanitize errors to exception class names so probe responses do not leak configured URLs or credentials.

## Validation

- `python -m pytest api/tests/unit/routers/test_health.py -q` -> 10 passed
- `docker compose -f docker-compose.yml config --quiet` with dummy required secrets
- `docker compose -f docker-compose.test.yml config --quiet` with dummy required secrets

## Notes

Readiness intentionally covers only core API-serving dependencies: Postgres, Redis, RabbitMQ, and configured S3-compatible storage. Worker and scheduler heartbeats are left out so API rollout is not coupled to optional or scale-dependent background processes.


jackmusick:
Thanks for this — solid work, well-scoped, and the credential-sanitization on probe errors is a nice touch. Queued auto-merge.

If there's more you wanted to do here (worker/scheduler heartbeats, additional endpoints, etc.), feel free to open another PR — happy to review.

Issue #74, opened 2026-04-25
[chore]: code-fix sweep for react-hooks 7.1 / eslint 10 expanded recommended set


jackmusick:
## Background

`eslint-plugin-react-hooks@7.1` and `eslint@10` ship an expanded recommended preset that adds:

- `react-hooks/set-state-in-effect` — `setState` calls in effect bodies (~40 sites)
- `react-hooks/refs` — ref access during render (~6 sites)
- `react-hooks/immutability` — direct mutation of state (~3 sites)
- `react-hooks/incompatible-library` — hooks called in incompatible contexts (~2 sites)
- `react-hooks/error-boundaries` — bad error boundary patterns (~1 site)
- `@eslint/js`'s `preserve-caught-error` — rethrows that lose `cause` (~8 sites)
- `no-useless-assignment` (~1 site)

Total: ~60 lint errors across ~30 files.

These are real signals, not noise — silencing them in eslint config would be a regression.

## Why this is being filed now

PR #72 (the npm minor-and-patch group bump) included `eslint-plugin-react-hooks` 7.0.1 → 7.1.1, which surfaced these errors and broke its CI. PR #73 added an ignore rule pinning the plugin at 7.0.x until this sweep lands.

PRs #61 (eslint 10) and #65 (typescript 6) were also closed for related reasons — both blocked on upstream peer-dep tightening.

## Scope

- Fix all current call sites (the list will be re-derived once we install plugin 7.1+ in a worktree)
- Where a `setState` in an effect is intentional (e.g., syncing external state), use the `set-state-in-effect`-suppressed pattern documented in the plugin (typically `useSyncExternalStore` or a callback ref)
- Where a `setState` is a code smell (cascading render), refactor to derive the state from props or memo
- For `preserve-caught-error`, add `cause` to rethrown errors
- For `react-hooks/refs`, move ref reads into effects or event handlers
- After fixing, remove the ignore rule for `eslint-plugin-react-hooks` and `eslint-plugin-react-refresh` from `.github/dependabot.yml` and let Dependabot land the bumps

## Estimated effort

~1 day of focused work. The fixes are mostly mechanical per category but require reading each effect to confirm the pattern.

## Acceptance

- `npm run lint` passes with `eslint-plugin-react-hooks@^7.1` and `eslint-plugin-react-refresh@^0.5` installed in a worktree
- Eslint 10 also passes (verify in same worktree by bumping `eslint` and `@eslint/js` to 10)
[...]

PR #86, opened 2026-04-25
fix(client): code sweep for eslint 10 + react-hooks 7.1 (gates #74)

jackmusick:
## Summary

Per issue #74. Lands the code-side fixes needed before bumping
eslint to v10 + eslint-plugin-react-hooks to v7.1 — the new
recommended set adds five rules that surface ~60 real, fixable
issues across the client.

**No eslint-disables were added.** All 60 errors are real refactors.
Five randomly-sampled diffs were re-eyeballed and they're all
genuine improvements (derived state instead of useEffect+setState,
useMemo replacing state mirrors, dead debug refs removed, etc.).

## Counts per rule

| Rule | Count | Typical fix |
|------|-------|-------------|
| `react-hooks/set-state-in-effect` | 41 | derive (`useMemo` / inline) or "adjust state during render with a prev-value sentinel"; async fetches wrapped in void-IIFE |
| `react-hooks/refs` | 6 | move ref reads into effects/event handlers; portal container converted from RefObject to state-via-callback-ref |
| `react-hooks/immutability` | 3 | hoist `refreshTokenInternal` / `loadMFAStatus`; use `window.location.assign(...)` |
| `react-hooks/incompatible-library` | 2 | extracted `CheckboxField` sub-component using `useWatch`. Two `form.watch(...)` call sites in `AgentSettingsTab` left as warnings (not errors) — converting them broke downstream `useMemo` hooks via `preserve-manual-memoization` |
| `react-hooks/error-boundaries` | 1 | split eval step out of `try/catch` so the rendered JSX is never inside a try block |
| `preserve-caught-error` | 8 | add `{ cause: e }` to rethrows in `passkeys.ts` / `useWorkflowMutation.ts` |
| `no-useless-assignment` | 1 | drop pointless `= false` initializer in `KeyboardContext` |
| **Total** | **60 errors fixed** | + 2 warnings remaining (informational, do not gate) |

## Top files touched

- `src/components/editor/RunPanel.tsx` (170 lines changed — `detectedItem` state → `useMemo`, dead debug ref removed, completion-cleanup wrapped in `queueMicrotask`)
- `src/components/forms/FormRenderer.tsx` (97 lines — extracted `CheckboxField` sub-component using `useWatch`)
- `src/components/editor/PackagePanel.tsx` (83 lines)
- `src/contexts/AuthContext.tsx` (82 lines — `refreshTokenInternal` hoisted + wrapped in `useCallback`)
- `src/pages/ExecutionDetails.tsx` (69 lines — portal container converted from `RefObject` prop to state; `signalrEnabled` derived during render)
- `src/pages/FormBuilder.tsx` (69 lines — adjusting-state-on-prop-change idiom)
- `src/components/forms/FormRenderer.tsx`
- `src/pages/settings/LLMConfig.tsx`
- `src/pages/settings/GitHub.tsx`
- `src/pages/ExecutionHistory.tsx`

## Other notes

- Bumps `client/tsconfig.app.json` `target` / `lib` from ES2020 to ES2022 so the new `Error(msg, { cause })` constructor typechecks. Root `tsconfig.json` already targets ES2022.
- Manifest deps (eslint, @eslint/js, eslint-plugin-react-hooks, eslint-plugin-react-refresh, typescript-eslint) NOT bumped here — they'll follow in a separate PR after the dependabot ignore rules are lifted, per issue #74's plan.

## Test plan

- [x] `npm run lint` passes with eslint 10 + react-hooks 7.1 installed locally (0 errors, 2 warnings — both pre-existing `form.watch` call sites)
- [x] `npm run lint` still passes against current eslint 9 (unchanged)
- [x] `npm run tsc` passes for production code (test files have pre-existing `@testing-library/react` type issues unrelated to this PR)
- [x] `./test.sh client unit` passes — all 763 vitest tests pass
- [ ] @jackmusick — eyeball welcome, especially the `set-state-in-effect` refactors. The "adjusting state during render with a prev-value sentinel" pattern is the React-recommended idiom (https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) but it's worth a sanity check on the bigger sites: `WorkflowEditDialog`, `ExecutionHistory`, `GitHub.tsx`, `LLMConfig.tsx`, `ExecutionDetails.tsx`.

Closes parts of #74 (gates the dep bump).

### DIFF
--- a/client/src/components/app-builder/EmbedSettingsDialog.tsx
+++ b/client/src/components/app-builder/EmbedSettingsDialog.tsx
@@ -121,10 +121,13 @@ export function EmbedSettingsDialog({
     }
   }, [appId]);
 
+  // Network fetches: setState happens after `await`, so wrapping in a void
+  // IIFE keeps the synchronous part of the effect free of setState calls.
   useEffect(() => {
-    if (open) {
-      fetchSecrets();
-    }
+    if (!open) return;
+    void (async () => {
+      await fetchSecrets();
+    })();
   }, [open, fetchSecrets]);
 
   // ========================================================================
--- a/client/src/components/applications/AppReplacePathDialog.tsx
+++ b/client/src/components/applications/AppReplacePathDialog.tsx
@@ -153,9 +153,13 @@ function FolderPicker({
 		}
 	}, []);
 
-	// Load root on mount
+	// Load root on mount. setState inside loadPath only fires after the
+	// awaited fetch resolves — wrapping in a void IIFE keeps the synchronous
+	// effect body free of setState calls.
 	useEffect(() => {
-		loadPath("");
+		void (async () => {
+			await loadPath("");
+		})();
 	}, [loadPath]);
 
 	const toggle = useCallback(
@@ -178,9 +182,12 @@ function FolderPicker({
 	);
 
 	// Auto-expand ancestors of the selectedPath when it changes (so the text
-	// field → tree sync works). Only expand, never collapse.
-	useEffect(() => {
-		if (!selectedPath) return;
+	// field → tree sync works). Only expand, never collapse. Adjust during
+	// render with a previous-selectedPath sentinel rather than via
+	// setState-in-effect.
+	const [prevSelectedPath, setPrevSelectedPath] = useState(selectedPath);
+	if (prevSelectedPath !== selectedPath && selectedPath) {
+		setPrevSelectedPath(selectedPath);
 		const parts = selectedPath.split("/");
 		const ancestors: string[] = [];
 		for (let i = 1; i < parts.length; i++) {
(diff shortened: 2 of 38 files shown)


jackmusick:
@jackmusick — eyeball welcome, especially the set-state-in-effect refactors. The prev-value-sentinel pattern is the React-recommended idiom for adjusting state when a prop changes (https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes), but the bigger sites (WorkflowEditDialog, ExecutionHistory, GitHub.tsx, LLMConfig.tsx, ExecutionDetails.tsx) are worth a sanity check.

Not enabling auto-merge — this PR genuinely needs review before landing.

PR #99, opened 2026-04-26
docs(spec): external MCP client (RFC)

sdc53:
Design doc for adding MCP-client capability to Bifrost — the symmetric counterpart to the existing `mcp_server/` package.

Filing as docs-only so the conversation has a stable URL to anchor on. Companion feature issue will be filed referencing this doc.

**TL;DR:** Reuse `integrations` + `oauth_providers` + `configs`, add a `kind` discriminator and one new `external_mcp_tools` cache table. `resolve_agent_tools()` gains a fourth source. Two-layer access control: Bifrost grants AND remote scopes.

First question in the spec is whether this is already on your roadmap — answer that first if so and I'll align rather than duplicate.

If green-lit I'd build the whole thing as a single feature PR (open question if you'd prefer it split, noted at the end of the spec).

### DIFF
--- a/docs/superpowers/specs/2026-04-25-external-mcp-client-design.md
+++ b/docs/superpowers/specs/2026-04-25-external-mcp-client-design.md
@@ -1,0 +1,87 @@
+# External MCP Client (RFC)
+
+**Date:** 2026-04-25 · **Status:** Draft
+
+Bifrost serves MCP via `mcp_server/`. This proposes the symmetric **client** capability: register external MCP servers as integrations and let Bifrost agents call their tools.
+
+## First question: is this already on your roadmap?
+
+Asking before describing the design — if you've already got something in flight or have a different shape in mind, point me at it and I'll align rather than duplicate.
+
+## Problem
+
+`resolve_agent_tools()` in `agent_helpers.py` builds an agent's toolset from three sources: system tools, workflows, delegated agents. All in-process. There's no path to call a tool hosted on an external MCP server.
+
+The MCP ecosystem has produced a lot of useful servers in the last six months (M365 Admin, Slack, Atlassian, Salesforce, plus vendor and homegrown ones). An operator today has to mirror each as Bifrost workflows — which doesn't scale, a 77-tool server is 77 mirrored workflows — or skip Bifrost and use Claude Desktop / Code directly. SpireTech has a concrete forcing example: an MCP server fronting HaloPSA with granular per-integration permissions. We want Bifrost agents to call it without standing up a parallel agent stack.
+
+## Design
+
+**Data model — extends existing tables, no parallel infra:**
+
+```sql
+ALTER TABLE integrations
+  ADD COLUMN kind VARCHAR(50) NOT NULL DEFAULT 'standard';
+-- 'standard' (today's behavior) | 'external_mcp'
+
+CREATE TABLE external_mcp_tools (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    integration_id UUID NOT NULL REFERENCES integrations(id) ON DELETE CASCADE,
+    name VARCHAR(255) NOT NULL,
+    qualified_name VARCHAR(511) NOT NULL,    -- e.g. 'halopsa.list_tickets'
+    description TEXT,
+    input_schema JSONB NOT NULL,
+    last_seen_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    UNIQUE (integration_id, name)
+);
+```
+
+OAuth credentials reuse `oauth_providers`. Base URL goes in `configs`. Per-agent grants either reuse `agent.tools` if it can be made polymorphic, or add a small association table.
+
+**Code surface (~1.5k LOC):** new `api/src/services/mcp_client/` (Streamable-HTTP client wrapper, catalog sync, dispatch, registration validation), a new ORM + alembic migration, ~30 lines added to `resolve_agent_tools()` for a fourth source, ~40 lines in `agent_executor.py` to branch dispatch on tool kind, new endpoints under `/integrations` for catalog refresh, plus `ExternalMcpForm.tsx` and a tweak to `AgentToolPicker.tsx`.
+
+**Run-time flow:** LLM emits `tool_use` with `name="halopsa.list_tickets"`, executor sees the qualified-name prefix, dispatches via the MCP client (which gets a token from `oauth_provider`'s existing client_credentials helper, reuses it until expiry, posts the tool call), result is normalized to the same envelope as a workflow tool result. Errors become tool-call errors, not fatal run errors.
+
+**Audit:** each call logs `(agent_run_id, integration_id, tool_name, status, duration)` via the same path workflow tools use, with `tool_kind="external_mcp"`.
+
+## Access control — two layers, either can deny
+
+| Layer | Question | Where |
+|---|---|---|
+| Bifrost | Can THIS agent invoke THIS tool? | per-agent grant, checked at `resolve_agent_tools` time |
+| Remote MCP server | Can THIS credential perform THIS scope? | the remote's own policy — returns 403 / tool error |
+
+A misconfiguration on either side is caught by the other. The remote server is treated as untrusted (responses validated against declared schema, size capped, timeouts enforced) but its tool definitions are trusted for naming/schema.
+
+## Non-goals
+
+Streamable HTTP only (no SSE/stdio). No tool-call streaming back through the LLM. No per-user OAuth-on-behalf-of (integration credentials are shared across agents that use it, same as today). Bifrost is a consumer, not a registry/proxy of other MCP servers.
+
+## Security highlights
+
+Encrypted-at-rest credentials reuse `oauth_providers.encrypted_client_secret` — no new crypto path. Token refresh via the existing `oauth_token_refresh` scheduler. Optional `MCP_CLIENT_ALLOWED_HOSTS` env var to restrict outbound destinations. Default 256 KB result cap, 30 s wall-clock per call (overridable per-integration via `configs`). Touches the executor, agent helpers, and audit log → expects manual sensitive-paths review per CONTRIBUTING.md.
+
+## Migration / compat
+
+`kind` defaults to `'standard'`, existing integrations unchanged. `external_mcp_tools` is purely additive. API responses gain a `kind` field; clients that ignore unknown fields are unaffected. Reversible.
+
+## Alternatives considered
+
+- **Mirror each remote tool as a workflow.** Doesn't scale; remote tool changes become workflow PRs.
+- **Single dispatcher tool** (`mcp_call(server, tool, args)`). Collapses N tool schemas into one, LLM can't reason about specific tools at planning time, breaks per-tool grants.
+- **Delegate to Anthropic's `mcp_servers=` parameter on `messages.create`.** Requires public exposure of every MCP server, locks tool dispatch into Anthropic's infra, bypasses Bifrost's grant gate. Could be a future per-integration fall-back option for operators who don't want to host the client.
+
+## Open questions
+
+1. `kind` column shape — plain VARCHAR vs normalized table vs JSON metadata. I'd default to plain.
+2. Per-org vs per-deployment registration — `integration.organization_id` (existing) seems right but want to confirm for multi-tenant deployments.
+3. Catalog refresh — manual button + ~24h scheduled, per-integration or global?
+4. Polymorphic `agent.tools` vs new association table — depends on whether the existing `Tool` model already wants a discriminator.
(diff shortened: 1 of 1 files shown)


sdc53:
Companion feature request: jackmusick/bifrost#100

jackmusick:
**Edit (cleaning up my stream-of-consciousness reply):** Trying to land the actual architectural concern more clearly.

This was on my roadmap and I like the data-model approach. Before I green-light, I want to surface a risk that I think is the real blocker, and ask for your take on it.

### What I think the actual value is

Reading this back, I don't think the value of this RFC is "MCP-the-protocol" — it's **catalog reuse**. Mirroring 77 HaloPSA tools as 77 workflows doesn't scale, and an MCP-aware client lets us inherit a vendor's catalog without that toil. That part I want.

### The risk I keep landing on

Bifrost integrations are **system-first**: org-scoped or global, headless agents, scheduled runs, no UI primitive for "your token expired, please re-consent." That model assumes credentials are configured once and refresh in the background.

Most useful vendor MCP servers (Slack, GitHub, M365 Admin, Atlassian, Notion) are **user-first**: per-user OAuth, interactive consent, the protocol leans into elicitation/sampling/etc.

Looking at how Copilot Studio, M365 Copilot Chat, Claude.ai, Cursor, and VS Code handle this — there's a remarkably consistent pattern: **admin enables the MCP server at the org level, then each user OAuths individually**. None of them have a "this tool is interactive-only" flag; tools are uniformly callable and the runtime just 401s if the wrong auth is attached. Headless only works with service-principal/managed-identity credentials.

So the architectural question I keep running into: **if we build the service-account-only version of MCP-as-client now, how do we extend it to per-user later without a refactor?** Specifically:

1. **Single point of configuration.** We'd want admins to configure external MCP servers once (URL, scopes, allowed tools), the same way they configure integrations today. How do you envision a user-auth tier overlaying that config without a separate registration path?

2. **Gating user-MCP from system contexts.** If a tool requires user auth, what happens when a scheduled agent or webhook-triggered agent tries to use it? Is the right answer "tools have an auth-context tag and the dispatcher refuses to bind them in non-interactive flows" — or do we just expect operators to know the rules and accept runtime 401s?

3. **Or: a third scoping layer.** Today integrations scope is `global → org`. A future where it becomes `global → org → user` would let a user-scoped MCP tool override an org-scoped one in chat contexts only. That feels right philosophically (matches the two-stage gate pattern in Copilot/Claude/Cursor) but it's a meaningful refactor of the integration model. **I'd genuinely value your thoughts on this** — does that scoping model make sense as a forward path, and what would the disposition rules be?

### What I'm asking

If you can sketch even a rough answer to (1) and (2) — or push back on (3) — I'd green-light. The thing I want to avoid is building a system-first MCP client now that we'd architecturally regret when (not if) the per-user MCP requirement shows up. If we agree on the auth-tier model upfront, the implementation can ship in stages.

Genuinely want to collaborate — the design itself is solid, I just want to make sure we build it on a foundation that survives the user-auth requirement we're definitely going to face.

sdc53:
Thanks — these are genuinely the right questions and I had to back up and rethink. You're right that lumping MCP into "always service-account" hides a real use case I have on the roadmap.

**Real forcing example:** SpireTech's Tech Support agent needs to combine HaloPSA results (service-account — halopsa-mcp's job) with SharePoint Copilot Graph search results, where Microsoft Graph filters results by the calling user's permissions. Using a service principal there is either useless (no SharePoint membership) or scary (Application permissions that bypass per-document ACLs). The whole point of Copilot search is the personalized result ranking — it needs the calling user's identity. So per-user delegation isn't a v2 nice-to-have; it's required for the integration to be correct.

**Proposed: admit both auth modes as configuration on the integration**, not as separate features:

- `auth_mode = 'client_credentials'` — admin sets up one credential per integration; every agent run uses it. Works in chat / scheduled / webhook contexts equally. (halopsa-mcp case.)
- `auth_mode = 'authorization_code'` — admin sets up the OAuth app once; each user does a one-time consent at `/integrations/{id}/connect`; per-user refresh tokens stored in a new `user_external_mcp_credentials` table. (SharePoint case.)
- *Optional, opt-in per integration:* `service_user_id` — a designated Bifrost user whose tokens are used as the fallback when the integration's `auth_mode=authorization_code` but the agent run has no caller (schedule/webhook). Disabled by default to avoid the "use the admin's super-token" foot-gun. Useful when you want an automated agent to query M365 as a dedicated service identity that has its own license + scoped permissions, rather than re-using a real user's token.

**The four concerns become:**

**1 & 2 (admin-vs-user, scheduled agents):** solved at the integration level. Service-mode tools always work. Per-user-mode tools require user context — `resolve_agent_tools()` filters them OUT when the agent run has no caller, *unless* the integration has a `service_user_id` configured. Avoids the anti-pattern: the agent's tool list reflects what auth context is actually available.

**3 (mid-chat re-auth):** solvable when Bifrost owns the chat UI. Dispatch returns a structured `needs_reauth` error with a `reauth_url`; the chat UI renders an inline reconnect button. Microsoft's refresh-token flow handles the silent path for non-expired refresh tokens. **You're right this does NOT solve the case where a third-party chat client uses Bifrost as an MCP server and Bifrost is delegating to another external server** — the third-party chat has no language for "this tool's downstream needs reauth." That genuinely is a v2 problem (or maybe never), but it's a separate deferral from the per-user delegation itself: per-user works fine for Bifrost-owned chat surfaces and for Bifrost agents driven from any context with a known user identity.

**4 (philosophical):** the SharePoint case is concrete proof that there's value beyond chat. A scheduled-or-chat-driven agent that needs to query M365 (or Salesforce, or any other vendor that ships an MCP server with delegated auth) gets to do it without us reimplementing search-with-permissions ourselves. That's not a "chat-client niche."

I'd plan to update the spec PR with this — the `auth_mode` column, the `user_external_mcp_credentials` table, the optional `service_user_id`, the executor's user-context plumbing, and the tool-availability filter for non-chat agent runs. The non-trivial bit is making `agent_executor` pass `current_user_id` through to dispatch — that touches the existing tool-dispatch contract, so worth flagging as a sensitive-paths change up front.

Per-user delegation needs to be in v1 in some form (full per-user, service-user fallback, or both — happy to follow your preference on which). Deferring it to v2 would mean the SharePoint integration class can't ship at all, and from what I'm seeing across the vendor MCP ecosystem (M365, Atlassian, Salesforce — all delegated by default) that's a significant chunk of the value.

Open to:
- Both per-user AND service-user (most flexible; users get personalized results in chat, schedules still work)
- Per-user only (cleanest; scheduled agents simply can't use SharePoint-class tools)
- Service-user only (simpler implementation; loses the personalized-results benefit in chat)

Strong preference any direction?

jackmusick:
Worked through this from a different angle and want to push back on the shape, not the goals.

**Don't put this on `integrations`.** The only thing it actually shares with a standard integration is OAuth. `integrations.get()` doesn't apply (MCP OAuth resource indicators differ from the underlying API's, so the credential isn't reusable outside MCP). Configs don't apply (the server URL is the integration, not a knob). Entity ID and data provider don't apply (multi-tenancy lives inside the protocol). `kind`-on-Integration means hiding most of the row most of the time, and every future Integration change has to ask "does this apply to MCP-kind too?" forever.

New top-level entity, own tab. Reuse is service-layer (`oauth_provider.py`, the OAuth provider editor widget extracted into a shared component), not schema-layer. `OAuthProvider.integration_id` is already nullable so this works without changes there.

**Drop the service slot.** Looked at the actual ecosystem — M365, Notion, Atlassian, GitHub, Halo's own Copilot Studio docs — vendor MCP servers are universally `authorization_code`. None implement `client_credentials`. Copilot's three OAuth modes (Discovery / Dynamic / Manual) are all delegated; Manual just means "the server's discovery is broken, paste URLs by hand."

Autonomous-run support comes from a *shared* delegated connection: admin OAuths once (ideally as a dedicated `bifrost-service@company.com` account, not personal), those tokens become the fallback. Two independent flags at connect time:

- ☐ Available in user chat (fallback for users who haven't connected)
- ☐ Available to autonomous agents (schedules, webhooks)

Both unchecked = personal use only; chat users without their own connection get prompted. Schema should be shaped to admit a service slot later if `client_credentials` MCP servers actually emerge, but don't design for it now.

The popup needs to state the consequence in user terms: *"users will read and modify resources visible to {your name}'s account — recommended only for dedicated service accounts, not personal accounts."* Not "delegated credentials shared."

**Per-org-only, no global.** Vendor MCP servers are inherently customer-scoped. Global is a footgun. `organization_id IS NULL` cascade pattern stays available for the rare non-tenant case if it ever shows up.

**Catalog is per-connection, not per-server.** Most servers scope `tools/list` by granted scopes, and per-org URL overrides mean different orgs can point at different deployments. Per-server caching either shows tools that 403 at dispatch or misses tools that exist.

**Server template + per-org connection.** Server carries auth params (token_url, scopes, audience, redirect URL), no secrets. Connection carries client_id, client_secret, optional per-param overrides, the two flags. Discovery-first at server creation (`/.well-known/oauth-authorization-server`, `/.well-known/oauth-protected-resource`); falls back to manual fields when discovery fails or returns wrong values.

**Bifrost won't ship a curated catalog of templates.** Maintenance, curation, and versioning aren't businesses I want to be in. Server templates round-trip through the manifest like every other entity, so users can share their working configs without us owning them.

**Per-user tokens get their own table.** `oauth_tokens.user_id` exists but it's for SSO (Bifrost user signing into Bifrost), not delegated API tokens. Different lifecycle, different concern.

**Other things worth nailing:**
- Streamable HTTP only, via the MCP SDK's `streamablehttp_client`. No SSE, no stdio.
- Redirect URL is deterministic per Bifrost deployment, displayed at server creation, registered once in the vendor's app config. No save-then-go-back loop.
- Webhook-triggered agents are strict (autonomous = no per-user token lookup). Permissive signed-user-claim path is documented and deferred.
- Publish-time validation: warn unmissably when a chat-only tool is on a schedulable agent, don't block.
- Standard-Integrations UX cleanup (org list without data provider, per-org overrides as first-class) is a separate change and isn't blocking this.

If the entity-not-discriminator reframe lands for you, please update the PR with the four tables (server, connection, per-connection catalog, per-user tokens), the auth resolution table, and the connect-popup wording. If you'd rather keep `kind`-on-Integration for some reason I haven't surfaced, happy to hear it — but the SDK / config / entity-id arguments all point pretty hard at "this isn't an integration."

Once we're aligned on the spec, fine with this shipping as a single feature PR. Executor user-context plumbing is the right place to flag for sensitive-paths review.

sdc53:
Reframe lands. Pushed back internally and couldn't — the SDK / config / entity-id arguments are hard to dispute, and once server-template and per-org-connection are separate concepts, trying to fit both into a `kind`-discriminated `integrations` row is where my original spec was leaning on the discriminator hardest.

Concretely agreeing with: new top-level entity, drop `client_credentials`, two-flag connection model + shared `bifrost-service@` account as the autonomous fallback, per-org connections only, per-connection catalog, server template + connection split, per-user tokens in their own table, streamable HTTP only, no curated catalog of templates.

I mocked up the screens (sidebar placement, MCP Servers list, server template + discovery form, connection edit with the two flags + the popup wording per your text, per-user My Connections, agent tool picker with auth-context badges, chat reauth, auth-resolution flow, schema sketch) to validate end-to-end. Three things came up while drawing it that I'd want to nail down before code lands — they can go in the v2 spec, not a separate round-trip:

1. **Server template scope.** Platform-level row, or per-org with `NULL = visible-to-all` (cascade pattern)? Manifest round-trip is cleaner with platform-level; cascade is cleaner if customers will fork templates per-org. Either works.
2. **Per-user token row.** Thin FK to `oauth_tokens`, or its own row carrying consent metadata (last-consent, granted scopes, reauth deadline)? The chat reauth UX needs "expiring soon" semantics, which argues for the richer row.
3. **URL override location.** Default URL on the template + optional override on the connection (my read), or set always at connection? Your text confirms per-org overrides; just want to lock whether the template carries a default.

**Status / handoff:** no code yet, just spec analysis. This is core Bifrost surface that you'll live with longer than I will — I'm genuinely happy to defer to you on driving the implementation if you'd prefer. Let me know either way.

Independently, I have downstream Copilot / SharePoint agent work I'd like to start tomorrow regardless of who picks up the upstream piece, so timing-wise: if you want to drive this, a quick yes/no helps me sequence around it. If I don't hear back I'll post a v2 of the spec per your reframe in the next day or two — including my best-guess answers to the three above — and pause before any code so the schema can lock first.

sdc53:
v2 spec posted as #176 — adopts your reframe, answers the three open questions, flags one additional decision (client_credentials v1 support for the halopsa-mcp forcing example) for your input. Starting implementation tomorrow regardless of review timing per the prior comment; schema can still adjust before code lands if you flag anything.

jackmusick:
Closing as superseded by #176 (the v2 spec, merged 2026-05-03) per your own note in both #176 and #177. The historical RFC stays preserved in the PR record; substantive review continues on #177.

PR #107, opened 2026-04-26
[codex] Add ZAP baseline scan

MTG-Thomas:
## Summary

- adds a weekly/manual OWASP ZAP passive baseline workflow for the exposed Bifrost POC endpoint
- keeps scan scope hard-coded to sanctioned targets only, with no arbitrary URL input
- documents the current scan posture, operating notes, and next hardening steps in `docs/security/ZAP-BASELINE.md`

## Validation

- Parsed `.github/workflows/zap-baseline.yml` with Python/PyYAML
- Confirmed `https://20.9.81.122` returns `200 OK` over HTTPS from this workstation

## Notes

This mirrors the Bifrost Docs ZAP posture: passive baseline only, no issue auto-writing, and `fail_action: false` while the first artifacts are triaged. The current target is the Azure POC public IP; the workflow and docs should be updated together when a stable hostname replaces it.

### DIFF
--- a/.github/workflows/zap-baseline.yml
+++ b/.github/workflows/zap-baseline.yml
@@ -1,0 +1,70 @@
+name: OWASP ZAP Baseline
+
+on:
+  workflow_dispatch:
+    inputs:
+      target:
+        description: Sanctioned Bifrost endpoint to scan
+        required: true
+        default: all
+        type: choice
+        options:
+          - all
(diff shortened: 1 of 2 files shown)


MTG-Thomas:
Closing because this was opened against upstream by mistake. This should stay in the fork/project lane unless explicitly requested.

Issue #120, opened 2026-04-26
RFC: scan findings lifecycle primitive for external security signals


MTG-Thomas:
## RFC

This is a request for direction before treating the accompanying implementation as a normal feature PR.

Bifrost already has workflows, org-scoped storage, apps, and operator-facing execution history. One reusable capability that seems adjacent to that vision is a small platform primitive for recurring external observations: ingest results from an external scanner, track each run, deduplicate repeated observations, resolve observations that disappear in later runs, and let operators transition finding state.

The prototype branch uses Nuclei as the first concrete producer because its output is structured and repeatable. The intended platform question is broader than Nuclei:

> Should Bifrost have a first-class scan/finding lifecycle API for external security or compliance signals, or should this remain workspace-specific code built only on tables/workflows?

## Prototype shape

The draft implementation adds an admin-only FastAPI router under `/api/scans/*` that stores data in existing org-scoped table documents instead of introducing new database tables.

It includes:

- scan run creation and history
- Nuclei finding ingestion
- severity validation
- occurrence-key hashing
- re-alert suppression for already open or acknowledged findings
- resolve-on-absence behavior for scanned hosts
- incomplete-run handling
- findings filtering by state, severity, and template tag
- bulk finding state transitions

## Why this might fit upstream

- It is a reusable lifecycle primitive for workflow-produced or externally produced findings, not an MSP-specific ticketing flow.
- It gives Bifrost apps and workflows a consistent backend surface for recurring scan observations.
- It avoids schema/migration overhead for the first slice by using the existing table/document store.
- It leaves downstream policy decisions, PSA tickets, and reporting in workspace code.

## Open questions

- Should the API remain Nuclei-specific for an initial concrete slice, or be generalized first to `security_findings` / `scanner_findings`?
- Should this stay as a thin router over existing tables, or move to explicit contracts/service modules before merge?
- Is admin-only access the right initial permission model?
- Is resolve-on-absence appropriate as default behavior, or should it be opt-in per scan run?
- Does this belong in core Bifrost at all, or should it be demonstrated as a workspace integration first?

## Related prototype

I am opening a draft PR from `MTG-Thomas:codex/nuclei-scan-ingestion-lifecycle` as a concrete reference implementation, not as a request to merge without design feedback.

MTG-Thomas:
Draft prototype PR opened for discussion: #121

jackmusick:
[...]

PR #121, opened 2026-04-26
RFC: scan findings lifecycle API prototype

MTG-Thomas:
## RFC framing

This is a draft PR attached to #120. Please treat it as a concrete RFC prototype, not a ready-to-merge feature request.

The question for upstream is whether Bifrost should own a small lifecycle primitive for recurring external scan/security observations, or whether this kind of behavior should remain entirely in downstream workspace code.

## What this prototype does

- Adds an admin-only `/api/scans/*` router for scan runs and findings.
- Uses existing org-scoped table documents for storage, avoiding a new schema/migration in the first slice.
- Implements Nuclei ingestion as the first concrete producer.
- Tracks lifecycle behavior that plain JSON dumping does not cover: occurrence-key hashing, re-alert suppression, resolve-on-absence, incomplete runs, filtering, and bulk state transitions.
- Wires the router through the normal FastAPI router exports and app registration.

## What this is not

- Not an MSP-specific Halo/Autotask workflow.
- Not a vulnerability scanner bundled into Bifrost.
- Not a final assertion that the API should remain Nuclei-named.
- Not a request to merge without deciding whether the upstream shape should be generalized first.

## Design questions for maintainers

- Should this be generalized from `nuclei_scans` to a broader `scanner_findings` or `security_findings` API before merge?
- Should this stay backed by generic table documents for the first slice, or should core models/contracts/services be introduced up front?
- Is admin-only access the right initial permission boundary?
- Should resolve-on-absence be default behavior, or explicit per ingest request?
- Is this aligned with Bifrost's platform direction, or better kept as workspace-level code using existing tables/workflows?

## Validation

- [x] `python -m compileall api/src/routers/nuclei_scans.py api/src/main.py api/src/routers/__init__.py`
- [x] targeted router/model smoke test
- [x] app route registration smoke for `/api/scans/*`
- [x] `python -m pytest api/tests/unit/test_nuclei_scans_router.py -q`
- [x] `python -m ruff check api/src/routers/nuclei_scans.py api/src/main.py api/src/routers/__init__.py api/tests/unit/test_nuclei_scans_router.py`
- [x] `git diff --check`

Note: broad `python -m pytest api/tests/unit -q` currently stops during collection on Windows because existing process-pool tests import Unix-only `resource`; that occurs before this Nuclei test module runs.

### DIFF
--- a/api/src/main.py
+++ b/api/src/main.py
@@ -79,6 +79,7 @@ from src.routers import (
     form_embed_secrets_router,
     export_import_router,
     docs_router,
+    nuclei_scans_router,
     platform_workers_router,
     platform_queue_router,
     platform_stuck_router,
@@ -569,6 +570,7 @@ def create_app() -> FastAPI:
     app.include_router(form_embed_secrets_router)
     app.include_router(export_import_router)
     app.include_router(docs_router)
+    app.include_router(nuclei_scans_router)
     app.include_router(platform_workers_router)
     app.include_router(platform_queue_router)
     app.include_router(platform_stuck_router)
--- a/api/src/routers/__init__.py
+++ b/api/src/routers/__init__.py
@@ -56,6 +56,7 @@ from src.routers.form_embed_secrets import router as form_embed_secrets_router
 from src.routers.export_import import router as export_import_router
 from src.routers.docs import router as docs_router
 from src.routers.jobs import router as jobs_router
+from src.routers.nuclei_scans import router as nuclei_scans_router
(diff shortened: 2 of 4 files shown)


jackmusick:
Closing the prototype. The design discussion continues at #120 — short version: this should land as a Bifrost app/workflow library rather than a core router. Full reasoning there.

Thanks for the concrete prototype, it made the conversation much sharper than an abstract RFC would have.

PR #175, opened 2026-05-01
fix(client): static import map fixes Firefox 150 bundle load failures

Cory-Covi:
Production Firefox 150 fails to load every Bifrost app with "bare specifier was not remapped" because Firefox 150 ships multiple-import-maps support gated behind dom.multiple_import_maps.enabled (default off), enforcing the legacy single-map rule strictly. Any runtime <script type="importmap"> registered after the module loader has started is rejected. main.tsx's top-level imports run before initReactShim(), so the runtime map (post-#165 single map; pre-#165 two maps) is always too late in Firefox 150. Chrome 133+ ships the same multi-map feature on, which is why it merged the late map silently.

Move the platform import map from runtime JS injection to a static <script type="importmap"> in client/index.html. The map URLs point to small ESM stub files generated by a Vite plugin
(client/src/build-plugins/bifrost-module-stubs.ts) that re-export from globalThis.__bifrost_* keys populated by initReactShim() at boot.

User-declared deps continue to be resolved at runtime, but only when an app actually has them. In that case we lazy-load es-module-shims in shim mode (where late importmap-shim registration is allowed). Apps with no user deps load with zero polyfill cost.

A drift test (client/src/lib/platform-modules.test.ts) pins the contract between the registry, the shim, the static HTML map, and the stub-generator so future additions can't drift silently.

App bundler is unchanged. Existing S3 bundles still load. No SCHEMA_VERSION bump, no cache invalidation, no risk to deployed apps.


jackmusick (approved):
APPROVE — static import map for Firefox 150.

## What I checked

- **Drift test (`platform-modules.test.ts`)** — 15/15 pass. The test pins all four corners of the contract: registry validity, `stubUrlFor` filename stability, static importmap completeness + ordering before any module script, and the shim assigning every globalKey. This is the right safety net for a fix whose failure mode is silent (specifier resolution falling back).
- **Production build** — green; 7 stub assets emitted under `dist/__bifrost_modules/` (one per platform module). Stub bodies look correct: `globalThis.__bifrost_react` → `export default m.default ?? m; export const X = m["X"];` with a clear error if the global is unpopulated.
- **Full vitest** — 943/943 pass on the merged tip.
- **Conflict resolution on merge with main** — `BundledAppShell.tsx` had an import-list conflict with #184's `setDefaultAppScope` addition. Resolved by keeping #184's import and dropping the no-longer-used `platformScope` import (this PR removed `setPlatformScope()` entirely in favor of populating `globalThis.__bifrost_*` at boot via `initReactShim()`). Pushed as commit 313e03e5 on your branch.
- **Architectural fit** — moving from runtime blob-URL maps to static-HTML map is the right answer here. Apps with no user deps now pay zero polyfill cost. User-dep apps lazy-load es-module-shims and use shim mode for late importmap registration; the user-dep map duplicates the platform keys so shim-mode modules can resolve them too. Same import-map behavior cross-browser (Firefox 150 strict mode and Chrome multi-map both work).

## Notes

- The `REACT_EXTERNALS` constant continues to keep esm.sh from bundling its own React, preserving the single-instance invariant.
- `registeredUserDepMaps` Set guards against re-registering the same shim-mode map for the same dep set.

Merging.

PR #176, opened 2026-05-02
docs(spec): external MCP client v2 — locked after #99 reframe

sdc53:
v2 of the external-MCP-client design after your 2026-04-30 reframe. Supersedes #99.

## What changed from v1

- Not a `kind` discriminator on `integrations` — new top-level entity (`mcp_servers`) with server-template / per-org-connection split.
- Per-user delegated tokens (own table) for the user-first case (M365 Copilot); shared service tokens with two visibility flags (`available_in_chat`, `available_to_autonomous`) for autonomous + chat-fallback (halopsa-mcp).
- Per-connection tool catalog. Per-org-only connections. Streamable HTTP transport only.

## Answers to the three open questions

| Question | Answer | Why |
|---|---|---|
| Server template scope | Platform-level row (`organization_id` nullable, NULL = visible to all orgs) | Cleaner manifest round-trip; cascade pattern available as escape hatch |
| Per-user token row | Richer `user_mcp_credentials` row with `oauth_token_id` FK + consent metadata | Chat reauth UX needs "expiring soon" semantics |
| URL override | Default on template, optional override on connection | Matches your "per-org URL overrides" wording |

## One additional decision flagged for review

`client_credentials` support in v1. You said drop it because vendor MCP servers are universally `authorization_code`. The forcing example I mentioned in #99 (halopsa-mcp running alongside Bifrost in prod) uses `client_credentials` per-integration JWTs. The spec keeps both flows supported in v1 — `client_credentials` connections store the token in `oauth_tokens` with no per-user mode. Push back if you'd rather defer that.

## Status

Locked, ready for implementation. Per your "fine with this shipping as a single feature PR" — implementation PR to follow on a separate branch off this same fork. As I mentioned in #99, I'm not waiting on review here; will start implementation tomorrow regardless. Schema can still adjust if you flag anything before code lands.

Once this is approved or merged, #99 should close as superseded.

### DIFF
--- a/docs/superpowers/specs/2026-04-25-external-mcp-client-design.md
+++ b/docs/superpowers/specs/2026-04-25-external-mcp-client-design.md
@@ -1,0 +1,431 @@
+# External MCP Client (RFC v2)
+
+**Version:** v2
+**Date:** 2026-05-02
+**Status:** Locked, ready for implementation
+**Supersedes:** v1 (2026-04-25)
+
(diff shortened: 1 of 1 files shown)


jackmusick (approved):
Solid plan. Thanks!

Issue #148, opened 2026-04-29
manifest sync: integration UUID rewrite collides with FK CASCADE on integration_config_schema


sdc53:
## Summary

Manifest sync fails with a `UniqueViolationError` on `ix_integration_config_schema_unique_key` when an `Integration` row's UUID gets rewritten (existing-by-name match in `_resolve_integration`).

## Repro

1. Have an `Integration` in the DB with UUID **A** and name `"Foo"`, with config_schema rows referencing integration_id = A.
2. Update `.bifrost/integrations.yaml` so `"Foo"` has UUID **B** (different).
3. Run `POST /api/github/sync`.
4. Sync fails with:
   ```
   sqlalchemy.dialects.postgresql.asyncpg.IntegrityError: <class 'asyncpg.exceptions.UniqueViolationError'>
   duplicate key value violates unique constraint "ix_integration_config_schema_unique_key"
   DETAIL: Key (integration_id, key)=(B, <some_key>) already exists.
   [SQL: INSERT INTO integration_config_schema ...]
   ```

## Root cause

In `api/src/services/manifest_import.py` in `_resolve_integration`:

1. **Line ~1766–1780** — when `existing_by_name` matches, an `Upsert` is built that rewrites the integration's `id` from A to B. SQLAlchemy emits an `UPDATE integrations SET id = B WHERE id = A`.
2. **Postgres FK** on `integration_config_schema.integration_id` is `ON UPDATE CASCADE`. The cascade migrates schema rows from A → B *during the same statement*.
3. **Line 1786** — the cache lookup is keyed on `integ_id` (the manifest UUID, B), but **the cache was populated before the upsert ran**, so it has schema rows under A, not B. Lookup at B returns `{}`.
4. **Line 1796–1814** — the loop doesn't see existing rows under B, takes the `else` branch, and tries to `INSERT` a new schema row at `(B, key)`. But CASCADE already migrated rows there → unique constraint violation.

```python
# Line 1785-1793 — cache lookup uses the NEW integ_id
if cache is not None:
    existing_cs_by_key = dict(cache["integ_cs"].get(integ_id, {}))  # integ_id = B
else:
    existing_cs_result = await self.db.execute(
        select(IntegrationConfigSchema).where(
            IntegrationConfigSchema.integration_id == integ_id  # B
        )
    )
    ...
```

[...]

PR #177, opened 2026-05-02
feat: external MCP client — let agents invoke remote MCP server tools

sdc53:
Implements the external MCP client feature per spec PR #176 (which supersedes #99). Single feature PR per your "fine with this shipping as a single feature PR" preference.

Updated 2026-05-03 with bug fixes from running this in production for a few days, plus a per-agent connection-grant model that closes a real authorization gap I caught while testing.

## What it does

Lets Bifrost connect to remote MCP servers and surface their tools to agents. Symmetric to the existing `mcp_server/` package. Forcing examples: Microsoft 365 Copilot Graph (per-user delegated, SharePoint/email/calendar) and the standalone `halopsa-mcp-server` running alongside Bifrost in production (`client_credentials` per-integration JWTs, 83 HaloPSA tools).

## Shape

Adopts your 2026-04-30 reframe in full:

- **New top-level entity** (not a `kind` discriminator on `integrations`) — server template + per-org connection split
- **Per-org-only** connections; server templates can be platform-level (`organization_id IS NULL`)
- **Per-connection tool catalog**, populated from `tools/list` after a service-account OAuth
- **Two visibility flags** on each connection: ☐ Available in user chat / ☐ Available to autonomous agents
- **Per-user delegated tokens** in their own `user_mcp_credentials` table with consent metadata; falls back to a shared service token when admin enables the chat flag
- **Streamable HTTP only** via `mcp.client.streamable_http.streamablehttp_client` (no SSE, no stdio)
- **Discovery-first** OAuth setup using the server's `/.well-known/oauth-authorization-server` and `/.well-known/oauth-protected-resource`; falls back to manual fields when discovery fails

Plus one shape addition that came out of running it for a few days:

- **Per-agent MCP connection grants** (new since the original spec). Default-deny: agents only see tools from MCP connections explicitly bound to them. The previous "all org agents auto-receive all enabled tools" behavior turned out to be too coarse — registering a write-capable connection silently widened every agent's capabilities. Backfill grants existing org agents to existing connections so the upgrade isn't a behavior change for current rows; new agents start clean.

## Auth resolution (5-path table)

When an agent invokes an MCP tool the dispatcher resolves which token to send:

1. **caller_user_id present, user has personal credential** → `user_mcp_credentials` token (vendor sees the user)
2. **caller_user_id present, no personal credential, `available_in_chat=True`, service token healthy** → shared service token
3. **caller_user_id present, no personal credential, no fallback** → `NeedsReauthError` (chat surfaces inline Connect button)
4. **caller_user_id is None (autonomous), `available_to_autonomous=True`, service token healthy** → shared service token
5. **caller_user_id is None, flag off** → tool was filtered at planning, dispatch never runs (assertion)

Plus: for `client_credentials` flows, the planner skips chat-context paths the user can't recover from (no per-user OAuth path exists), and dispatch raises `MisconfigError` (admin to fix) instead of `NeedsReauthError` (user to fix). The user can't reauthenticate on a server-to-server flow.

## Schema (5 new tables)

- `mcp_servers` — template (URL, OAuth provider FK, redirect URL, discovery snapshot)
- `mcp_connections` — per-org (server FK, org FK, client_id/encrypted_client_secret, two flags, optional service_oauth_token_id)
- `mcp_connection_tools` — per-connection catalog
- `user_mcp_credentials` — per-user delegated tokens with consent metadata
- `agent_mcp_connections` — per-agent grant join table (new)

All upserts are by UUID, never by natural key — explicitly avoiding the integration upsert-by-name FK CASCADE bug filed as #148.

## What you'd want to focus on in review

- **Auth-context plumbing in agent executors** — `caller_user_id` is now threaded from the request → `enqueue_agent_run` → worker (Redis context) → both executors → `resolve_agent_tools()` → `_execute_tool()` → `mcp_client.dispatch.invoke()` → `auth_resolution.resolve_token()`. The autonomous executor previously dropped `_caller`; this PR makes it load-bearing.
- **OAuth state signing** — `mcp_client/oauth_state.py`. JWT (HS256, audience-bound to `bifrost-mcp-oauth-state`, ≤10min exp). Replay defense via Redis nonce single-use atomic DEL.
- **Manifest deletion semantics** — tools follow the manifest; user credentials are user-owned and never bulk-deleted by manifest sync (CASCADE on connection_id handles connection-removal cleanup).
- **Per-agent grant authorization** — `set_mcp_connection_grants` is gated by org write scope; non-admins are silently denied (matches existing agent ownership semantics).

## Notable decisions you didn't address

- **`client_credentials` v1 support kept** — the halopsa-mcp forcing example uses it. Schema treats it as a service token in `oauth_tokens` with no per-user mode. `OAuthProvider` ORM already supports both flow types so the schema cost is zero. Push back if you'd rather defer.
- **Server template scope** = platform-level by default (`organization_id` nullable, NULL = visible to all). Cleaner manifest round-trip than per-org default.
- **Per-user token row** is its own table with consent metadata, not a thin link to `oauth_tokens.user_id` (which is for SSO).
- **URL override** = default on template, optional override on connection.
- **Per-agent grants default-deny** with a one-time backfill of existing rows. Closes a real authorization gap rather than relying on operator discipline.

## Operational guardrails added since the spec

- **MCP tool result size cap** (250KB / ~60K tokens). Protects against vendor responses that would blow Claude's 1M-token context window. Replaces oversized payloads with a structured truncation message that nudges the LLM toward filtering parameters on retry.
- **Per-tool admin disable persists across catalog re-syncs** — when an admin unticks a tool, the row gets `disabled_reason="Manually disabled by admin"` so the next `tools/list` refresh doesn't auto-re-enable it.

## Backend gaps deferred

UI degrades gracefully without these; not blocking for v1:

- `GET /api/agents/{id}/triggers` (used to tighten the publish-time warning panel — currently permissive)

## Testing

- pyright clean on new + modified files
- ruff clean
- 1300+ unit + e2e tests passing across the manifest round-trip, all 5 auth resolution paths, OAuth state PKCE/replay, executor caller propagation, both auth flows, per-tool toggle persistence, per-agent grant filter, migration backfill replay, dispatch result-size cap
- TypeScript / ESLint clean

## Statistics

- 19 commits, ~10K lines added across backend + frontend + tests
- 5 ORM tables, 2 alembic migrations, 7 routers, 1 new service package, 5 admin/user pages, 2 chat components

## Out of scope (deferred)

- Third-party MCP-client connecting to Bifrost-as-MCP-server with signed user claims
- Curated catalog of server templates (manifest is the sharing mechanism)
- The triggers endpoint flagged above

Once this and #176 (the spec) are reviewed, #99 should close as superseded.


jackmusick (approved):
APPROVE — external MCP client.

## What I checked

**Architecture and code (read in detail):**
- `auth_resolution.py` — 5-path token resolution (USER_TOKEN / SERVICE_FALLBACK_CHAT / SERVICE_FALLBACK_AUTONOMOUS / NEEDS_REAUTH / MISCONFIG); client_credentials handling raises MisconfigError correctly because there's no per-user OAuth recovery for server-to-server flows; refresh-in-place reuses scheduler primitives; resolution path enum recorded for audit.
- `oauth_state.py` — JWT state with HS256, audience-bound (`bifrost-mcp-oauth-state`), 10-min TTL, single-use nonce via atomic Redis DEL, PKCE S256 challenge.
- `agent_helpers.resolve_agent_tools` — org-scoped via `agent.organization_id`; planner-time autonomous gate filters connections without `available_to_autonomous` AND a usable service token; `caller_user_id=None` skips per-user lookups; `mcp__<connection_uuid>__<tool>` qualified-name parsing validates UUID before routing.
- Both executors thread `caller_user_id` correctly: chat executor pulls from `conversation.user_id`, autonomous executor pulls from `_caller["user_id"]` with UUID validation, both pass through to `mcp_dispatch.invoke` → `auth_resolution.resolve_token`. Errors translated to ToolResult envelopes (chat surface) or ToolError (autonomous).
- Per-agent grant migration backfills `(agent_id × connection_id)` pairs in the same org with `granted_by=NULL`, idempotent via `ON CONFLICT DO NOTHING`. Default-deny for new agents.
- Manifest round-trip tests cover both server templates and connections (`TestMCPServerManifest`, 8 tests including byte-identical round-trip and secret-not-in-manifest).
- CodeQL findings addressed: `log_safe()` applied to user-controlled input in router log lines; clear-text-password log at `auth_resolution.py:113` replaced with the token-id-only variant in commit cbb4ad03.

**Local verification (this branch + my merge of main):**
- `./test.sh all` on the merged tip: **4844 passed, 43 skipped, 0 failed** in 18:55. JUnit at `/tmp/bifrost/test-results.xml`.
- `./test.sh client unit` (vitest): full run shows 3 transient failures in `ExecutionSidebar.test.tsx` and `AgentTuneWorkbench.test.tsx`; both files PASS in isolation (15/15 and 17/17), and PASS together with `--no-file-parallelism` (32/32). Pre-existing vitest parallel-runner state pollution, not caused by this PR — neither file is touched by the diff.
- `bifrost-init` migrates clean on the merged tip (init exit 0, "Database migrations completed successfully"). Down_revision repointed from `20260428_webhook_rate_limit` to `20260504_backfill_table_access` to unify the alembic chain after main's table-policies migrations landed.
- CI on the merged tip: green across Lint & Type, Unit, E2E, CodeQL python+JS.

## Notes

- The migration repoint was the only blocker I found. Two heads in alembic would have failed `bifrost-init` with "Multiple head revisions are present" on merge. Now unified.
- Autonomous run will only see MCP tools whose connection has `available_to_autonomous=True` AND a usable service token — planner-time hard filter, not a runtime fail-closed.
- Per-agent grants default-deny is the right choice; the backfill preserves pre-feature behavior for existing rows.

Merging.

jackmusick (approved):
APPROVE — external MCP client. See full review notes in comments. Local: 4844 backend passed / 0 failed, vitest passes per-file (3 parallel-runner flakes pre-exist on main). Merged tip: bifrost-init exit 0 on the rebased migration chain. CI green on the merged tip across Lint/Type/Unit/E2E/CodeQL.

PR #225, opened 2026-05-08
fix(mcp): emit array `items` so Gemini accepts system tool schemas

mlspivey:
## Summary

- Google AI Studio (Gemini) rejects function declarations whose array parameters omit `items`. OpenAI and Anthropic accept the field. The system-tools helper `python_type_to_json_schema` in `api/src/services/mcp_server/server.py` was emitting bare `{"type": "array"}`, which broke every Gemini-routed agent that has tools with list params (e.g. `create_agent.system_tools`, `update_agent.delegated_agent_ids`, `forms.fields`, `tables.columns`).
- Fix: recurse on `get_args(annotation)` for parameterized lists; fall back to `{"type": "string"}` for bare `list`. The change is purely additive — both other providers already accept the field, and the workflow-tools path (`tool_registry.py:219-221`) has been emitting `items` to OpenAI/Anthropic in production all along.
- Adds a regression test that scans every real `get_system_tools()` output for `type: array` schemas missing `items`, so a future system tool with a list param can't silently re-introduce the bug.

## Error this fixes

```
GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[system_tools].items: missing field.
GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[knowledge_sources].items: missing field.
GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[delegated_agent_ids].items: missing field.
... (9 total)
```

## Notes

- The fix doesn't address Gemini's separate intolerance of `additionalProperties` on object schemas. If a follow-up Gemini error surfaces after this lands, that's the next layer to peel back (likely a Gemini-specific schema sanitizer in the LLM client).
- No production behavior change for OpenAI- or Anthropic-routed agents.

## Test plan

- [x] `tests/unit/services/mcp_server/test_tool_schemas.py` — 11 tests pass (5 new targeted helper tests + 1 new high-value regression test asserting every system tool's array params declare `items`).
- [ ] Manually try Gemini route after merge to confirm the original 400 is gone.

### DIFF
--- a/api/src/services/mcp_server/server.py
+++ b/api/src/services/mcp_server/server.py
@@ -453,7 +453,18 @@ def get_system_tools() -> list[dict[str, Any]]:
                 schema["additionalProperties"] = True
             return schema
         elif annotation is list or origin is list:
-            return {"type": "array"}
+            # Gemini rejects array schemas without `items`; OpenAI/Anthropic
+            # accept the field, so emit it unconditionally.
+            array_schema: dict[str, Any] = {"type": "array"}
+            if origin is list:
+                args = get_args(annotation)
+                if args:
+                    array_schema["items"] = python_type_to_json_schema(args[0])
+                else:
+                    array_schema["items"] = {"type": "string"}
+            else:
+                array_schema["items"] = {"type": "string"}
+            return array_schema
 
         return {"type": "string"}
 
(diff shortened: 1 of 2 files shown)


jackmusick:
Can you test this locally with your OpenRouter key?

mlspivey:
Verified locally with my API key. Tested Haiku, Gemini, and Kimi with tool usage in a chat with an agent to verify success.

jackmusick (approved):

PR #229, opened 2026-05-09
Use SeaweedFS for local object storage examples

MTG-Thomas:
## Summary
- replace the bundled local MinIO services with SeaweedFS in Compose, dev, debug, and test examples
- keep the application contract S3-compatible by only changing example endpoints, credentials, and docs
- remove MinIO-specific bucket-init containers in favor of SeaweedFS bucket creation from environment variables

## Validation
- `docker compose -f docker-compose.yml -f docker-compose.dev.yml config --quiet`
- `docker compose -f docker-compose.test.yml config --quiet`
- `docker compose -f docker-compose.debug.yml config --quiet`
- `git diff --check`
- `./test.sh stack up` partially exercised the test harness: SeaweedFS image pulled, container started, credentials loaded, and bucket `bifrost-test` was created. The harness then hung while starting app-side test services, so I stopped the wrapper and cleaned up the test stack.


jackmusick (dismissed):

jackmusick (approved):
Verified locally: full unit suite (3764) and storage/manifest e2e (15) pass on SeaweedFS. Confirmed the files.py drain-loop and empty-prefix filter are necessary — reverting them breaks test_delete_folder_removes_children against SeaweedFS. Switch is well-justified given MinIO went into maintenance mode in Dec 2025.

PR #243, opened 2026-05-13
fix: reap completed process pool workers

MTG-Thomas:
## Summary
- reap raw forked process-pool children when they exit
- join completed on-demand workers before removing them from pool state
- avoid reporting a completed child as crashed when its result is already queued

## Testing
- `python3 -m py_compile src/services/execution/process_pool.py tests/unit/execution/test_process_pool.py`
- `PYTHONPATH=$PWD/api pytest api/tests/unit/execution/test_process_pool.py -q -k "PidWrapper or reaps_completed_child or queued_result or crash_detection_replaces_process or sigkilled_worker"` (7 passed)

Note: the full `test_process_pool.py` run reached 42/43 in my ad hoc WSL venv; the remaining failure is an existing admission-control test trying to resolve `redis:6379` without the compose test environment.

### DIFF
--- a/api/src/services/execution/process_pool.py
+++ b/api/src/services/execution/process_pool.py
@@ -1157,6 +1157,7 @@ class ProcessPoolManager:
                         await self._handle_result(handle, result)
                         continue
                 except Empty:
+                    # No queued result: handle this dead process as a crash below.
                     pass
 
                 # Case A: unexpected crash
--- a/api/src/services/execution/process_pool.py
+++ b/api/src/services/execution/process_pool.py
@@ -33,7 +33,7 @@ import subprocess
 import time
 import uuid
 from dataclasses import dataclass
-from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


MTG-Thomas:
Updated the PR to cover the clean-exit/result-delivery race that showed up in E2E. The new commit adds a short grace period for exit_code=0 workers whose result pipe has not delivered yet, while preserving the crash path after the grace period.

Verification:
- CI is green at 9c8556fb (lint/type check, unit, E2E, CodeQL).
- Local Docker E2E on a Linux test VM: 1235 passed, 49 skipped.

jackmusick (approved):
Reaping logic and grace-window are well-scoped. Verified locally: full unit suite (3777) passes on the branch, including the 5 new regression tests pinning the clean-exit + queued-result race and the zombie-reap path. Important fix for on-demand mode as we lean on it harder for Chat V2 sandboxes.

PR #272, opened 2026-05-21
Fix platform admin user creation

MTG-Thomas:
## Summary
- auto-select the provider organization when platform-admin user forms receive organization data after the user type is selected
- cover the delayed-organization load path in CreateUserDialog
- codify that Bifrost verification should route through an e2e VM on pve-t340 unless local-only testing is explicitly requested

## Verification
- Not run locally; repo guidance now requires Bifrost testing through an e2e VM on pve-t340 unless explicitly requested otherwise.


MTG-Thomas:
Closing: this was opened against upstream by mistake. Re-opening against the MTG-Thomas fork only.

PR #297, opened 2026-05-24
Add Azure Blob object storage backend

MTG-Thomas:
## Summary
- add an Azure Blob-backed object storage client alongside the existing S3 client
- add runtime provider selection with Blob config via `BIFROST_OBJECT_STORAGE_PROVIDER=azure_blob`
- preserve existing object-key/upload/download/signed-url semantics through shared storage contract coverage
- document the Azure Blob opt-in environment variables in the README

## Verification
- `git diff --check`
- `ruff check api/tests/contract/test_object_storage_contract.py api/tests/unit/services/test_file_storage_backend_selection.py`
- `PYTHONPATH=api BIFROST_SECRET_KEY=0123456789abcdef0123456789abcdef python -m pytest api/tests/unit/services/test_file_storage_backend_selection.py api/tests/contract/test_object_storage_contract.py -q`
  - result: 6 passed, 1 skipped

## Notes
- `api/tests/unit/routers/test_files_signed_url.py` still does not collect in this workstation Python environment because `aio_pika` is missing.


MTG-Thomas:
Follow-up pushed after live deployment validation: Blob mode now routes repo storage, startup/reindex listing, and /health/ready through the active Azure Blob backend instead of requiring a configured S3 endpoint. This avoids a hidden S3-compatible storage dependency when BIFROST_OBJECT_STORAGE_PROVIDER=azure_blob is selected.

Additional local verification:
- python -m ruff check README.md api/src/services/repo_storage.py api/src/services/file_storage/azure_blob_client.py api/src/services/file_storage/s3_client.py api/src/routers/health.py api/src/services/file_storage/reindex.py api/tests/unit/test_repo_storage.py api/tests/unit/routers/test_health.py api/tests/unit/services/test_file_storage_backend_selection.py -> passed
- python -m pytest api/tests/unit/routers/test_health.py api/tests/unit/test_repo_storage.py api/tests/unit/services/test_file_storage_backend_selection.py -q -> 34 passed

MTG-Thomas:
Having some issues with storage, will revise upstream PR once smoked out and resolved.

jackmusick:
Did you give up?

MTG-Thomas:
No, agent misfire.

MTG-Thomas:
@jackmusick - I had something like 60 worktrees on my machine that I'm cleaning up today--composer 2.5 took, "clean up anything stale from our fork" a bit too literally once or twice, including this branch.

jackmusick:
No, not Codex! Edit: Composer

MTG-Thomas:
Cursor mega-struggling with my vague language today.

MTG-Thomas:
This should be now legitimately ready for review @jackmusick

jackmusick (changes requested):
Good foundation here and I want to land it, but a few things are blocking and one of them bugs me.

First, the thing I had to go check: this CI run is green but it isn't testing Azure, or S3. The contract suite in `test_object_storage_contract.py` only adds the S3 and Azure backends to the param list when `BIFROST_STORAGE_CONTRACT_*` env vars are set (the `_storage_backends()` helper drops them with `return None` otherwise). CI doesn't set them, so the whole suite runs against the in-memory fake. So "contract tests pass" currently means "a dict behaves like a dict." We need either Azurite wired into CI, or a real Azure container the suite runs against — otherwise the Azure path has no actual coverage.

Second, even with real-backend tests, two subsystems would slip through because they never go through the abstraction:

- `app_storage.py` is untouched — the whole `_apps/` preview/publish path calls aiobotocore directly (create_client/copy_object/put_object/get_object/list_objects_v2). In Azure mode this still goes to S3.
- `app_bundler/__init__.py::_write_live()` reaches into `app_storage._get_client()`/`_bucket` and does `put_object` itself — same problem for live bundles, and it's poking AppStorageService's privates.

Both need to route through the provider-selected client, and the publish path needs a test under the real backend (the contract suite structurally can't catch these since they bypass it).

Last, a scope question: `git_repo_manager.py` shells out to `aws s3 sync` — no Azure equivalent. Fine if repo-manager is out of scope for Azure, but let's say so in a comment rather than leave it as a landmine.

Needs a rebase too (conflicts against main).

MTG-Thomas:
Updated this branch after the rebase and follow-up coverage pass. The maintainer feedback items should now be covered:

- CI wires SeaweedFS/S3 and Azurite/Azure Blob into the contract test runner.
- The contract suite exercises the storage adapter matrix with `[memory]`, `[s3]`, and `[azure-blob]` cases, including signed URL HTTP round trips.
- `_apps/` coverage now exercises `AppStorageService` against both real backends for preview publish, direct live writes, and `_repo` to preview sync.
- `app_bundler` live builds route through `AppStorageService.write_live_file` instead of private S3 internals.
- `GitRepoManager` now fails fast in Azure Blob mode before invoking `aws s3 sync`; the S3-only scope is both documented and covered by a unit test.

Current head: `a83da9bfe3a7605611b3e8c30c274f8b9e3fb4513`.

Additional VM verification on `pve-t340.netbird.cloud`, VM 101:

- `./test.sh tests/unit/services/test_git_repo_manager.py tests/contract/test_object_storage_contract.py::test_app_storage_publish_uses_selected_real_backend tests/contract/test_object_storage_contract.py::test_app_storage_live_write_uses_selected_real_backend tests/contract/test_object_storage_contract.py::test_app_storage_sync_preview_uses_selected_real_backend -v --durations=10` -> 23 passed, including `[azure-blob]` app-storage cases.
- `./test.sh tests/contract/test_object_storage_contract.py -v --durations=10` -> 20 passed, 1 skipped; the skipped case is the in-memory fake for HTTP signed URL round trip, while `[s3]` and `[azure-blob]` passed.

PR #357, opened 2026-06-08
fix: normalize passkey transports for webauthn

MTG-Thomas:
## Summary
- normalize stored passkey transport strings into `py_webauthn` `AuthenticatorTransport` enums before creating WebAuthn credential descriptors
- add regression coverage for the `options_to_json()` registration-options path

## Why
With `webauthn 2.7.1`, `options_to_json()` expects descriptor transports to be enum values and accesses `.value`. Existing persisted passkeys store transports as JSON strings like `"internal"`, so registering an additional passkey can fail while serializing `excludeCredentials`.

This also applies to targeted authentication options when `allowCredentials` is built from stored passkeys.

## Verification
- `python -m pytest api/tests/unit/services/test_passkey_service.py -q`
- `python -m ruff check api/src/services/passkey_service.py api/tests/unit/services/test_passkey_service.py`

### DIFF
--- a/api/src/services/passkey_service.py
+++ b/api/src/services/passkey_service.py
@@ -29,6 +29,7 @@ from webauthn.helpers import (
     parse_registration_credential_json,
 )
 from webauthn.helpers.structs import (
+    AuthenticatorTransport,
     AuthenticatorSelectionCriteria,
     PublicKeyCredentialDescriptor,
     ResidentKeyRequirement,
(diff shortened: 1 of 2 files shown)


jackmusick (approved):
✅ Recommend approve. Correct fix for the py_webauthn 2.7.1 `options_to_json()` regression — stored JSONB transport strings (`"internal"`) are now normalized to `AuthenticatorTransport` enums before building `PublicKeyCredentialDescriptor`, on both the registration (`excludeCredentials`) and authentication (`allowCredentials`) paths. Tolerant of empty/unknown values. Regression test exercises the real `options_to_json()` serialization that was failing. CI + E2E green. Touches an auth-sensitive path but the change is well-scoped and tested.

PR #419, opened 2026-06-29
Fix virtual import resolution edge cases

### LINKED ISSUE #418: First-party workflows needing platform internals: DB-access proxy + two virtual-import edge cases
## Context

We run a fleet of first-party workflows/data-providers inside Bifrost workers. Recently we hit three rough edges around the workflow execution model + the virtual import system. Filing together since they're related; happy to split or PR any of them.

### 1. A sanctioned DB-access path for first-party workflows

Execution children are (correctly) sandboxed from `BIFROST_DATABASE_URL`, with the intended model being "workflow code uses the SDK over HTTP." That works well for most things, but for apps with genuinely **relational** data (FKs, joins, aggregations) the only options today are:

- keep DB credentials in the child (defeats the sandbox), or
- hand-build a REST layer (router + repo + contracts) per entity in the platform.

It would help to have a **blessed, scoped, authenticated in-child query/ORM proxy** — conceptually mirroring the S3 "API endpoint fallback" — so first-party/platform-owned workflows can do scoped relational access without holding platform credentials. Org-scoping/auth would resolve from the execution context as the SDK calls already do.

### 2. `STDLIB_PREFIXES` allowlist in the virtual import finder is incomplete

`VirtualModuleFinder` (in `services/execution/virtual_import.py`) intercepts any top-level module name **not** in `STDLIB_PREFIXES` and tries to load it from the module cache. `httpx` and `anyio` are listed, but their transitive deps (`httpcore`, `h11`, `h2`, `hpack`, `hyperframe`, `socksio`, `trio`, `outcome`, …) and other common third-party libs (`azure`, `numpy`, `pandas`, …) are not. When a workspace module imports one of those at module level, the finder intercepts the transitive dep, attempts a cache fetch, and can hit a partial-initialization circular import.

Maintaining the denylist is fragile. Consider deriving "is this a workspace module?" from the workspace module **index** (return `None` immediately for anything not present) instead of an allowlist of everything-that-isn't-workspace.

### 3. `find_spec` precedence — platform files can shadow workspace packages

`VirtualModuleFinder` is appended to `sys.meta_path` (so it runs **after** the default `PathFinder`). If a platform module on the filesystem path shares a name with a workspace **package** (e.g. a platform `src/services/<name>.py` vs a workspace `<name>/` package), `import <name>` resolves to the platform file (a module, no `__path__`), and any `from .sub import …` inside the workspace package then fails with `'<name>' is not a package`.

Prepending the finder for known workspace names (or otherwise giving workspace packages precedence) would prevent this collision class. It's subtle to debug because it only manifests in the real worker execution context, not in isolated imports.

Thanks for Bifrost — happy to contribute PRs for (2)/(3) if the direction sounds right.

jackmusick:
## Summary
- gate virtual module fetches on the module index so third-party imports fall through without Redis/API/S3 probes
- install the virtual import hook before filesystem finders so workspace packages cannot be shadowed by same-named modules on sys.path
- add regression coverage for non-workspace third-party imports and virtual-package precedence

## Issue linkage
Refs #418. This PR addresses only the two virtual-import edge cases described there:
- non-workspace third-party imports should not probe Bifrost module storage
- Bifrost workspace packages should not be shadowed by same-named filesystem modules

It intentionally does not close #418; the DB-backed platform data / trusted internal API discussion remains open.

## Verification
- ./test.sh tests/unit/services/test_virtual_import.py -q
- ./test.sh tests/unit/services/test_virtual_import.py tests/unit/test_virtual_import_s3_fallback.py tests/unit/test_solution_import_root.py tests/unit/test_solution_module_isolation.py -q
- ./test.sh tests/e2e/platform/test_virtual_import_integration.py -q
- ./test.sh quality api
- ./test.sh unit -q
- ./test.sh tests/e2e/api/test_builtin_events.py::TestBuiltInWorkflowEvents::test_workflow_failure_and_retry_exhausted_events_run_subscribers -q
- ./test.sh all

Note: an initial ./test.sh all attempt exposed a stale local test image with /entrypoint.sh permissions baked incorrectly; rebuilding the worktree test stack fixed the worker container, and the targeted failing e2e plus full backend suite then passed.

### DIFF
--- a/api/src/core/module_cache_sync.py
+++ b/api/src/core/module_cache_sync.py
@@ -99,6 +99,11 @@ def _candidate_storage_paths(path: str) -> list[str]:
     return [rooted]
 
 
+def candidate_module_paths(path: str) -> list[str]:
+    """Storage paths that may contain a concrete module file."""
+    return _candidate_storage_paths(path)
+
+
 def candidate_index_prefixes(base_path: str) -> list[str]:
     """Storage-path prefixes to scan the module index with, for namespace-package
     (PEP 420) detection of ``base_path`` (e.g. "modules").
--- a/api/src/services/execution/virtual_import.py
+++ b/api/src/services/execution/virtual_import.py
@@ -32,6 +32,7 @@ from typing import Any
 
 from src.core.module_cache_sync import (
     candidate_index_prefixes,
+    candidate_module_paths,
     get_module_index_sync,
     get_module_sync,
     solution_has_submodules,
@@ -258,12 +259,12 @@ class VirtualModuleFinder(MetaPathFinder):
     """
     Meta path finder that loads workspace modules from Redis cache.
 
-    Converts Python module names to file paths and fetches directly
-    from Redis. Each import attempt does a Redis GET for the module path.
+    Converts Python module names to file paths, confirms they are workspace
+    modules via the module index, then fetches the source from Redis/API/S3.
 
     Key design points:
-    - No hardcoded prefix required - works with any module name
-    - Direct Redis fetch - newly-added modules are immediately available
+    - Module index is the source of truth for workspace module ownership
+    - Non-workspace third-party imports fall through to normal import handling
     - Supports both modules (.py) and packages (__init__.py)
     """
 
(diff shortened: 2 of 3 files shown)


Glinte:
Found this issue while I'm looking for import related issues to test my debugging tool. My AI found this implementation isn't a complete fix, with the counterexample of `__hello__` which is a standard frozen package. As you guys are already aware, maintaining the `STDLIB_PREFIXES` is probably not a good long term idea. Though from what I can tell it works fine and the issue here is extremely minor. AI reproduction: https://github.com/Glinte/metapathology/tree/main/reproductions

There seems to be two issues:

1. This comment is simply wrong.

https://github.com/gobifrost/bifrost/blob/9d8b1e865c02b3473e52928b4d226ebc349e502e/api/src/services/execution/virtual_import.py#L52

2. Inserting your finder at the front means your finder is also before `BuiltinImporter` and `FrozenImporter`. You guys could just locate the index of (the earliest instance of) `PathFinder`, then insert `VirtualModuleFinder` before `PathFinder`. (or alternatively move them before your `VirtualModuleFinder`)

https://github.com/gobifrost/bifrost/blob/9d8b1e865c02b3473e52928b4d226ebc349e502e/api/src/services/execution/virtual_import.py#L452-L455

PR #479, opened 2026-07-14
Enable loose v2 app migration staging

jackmusick:
## What changed

- allow authored loose `standalone_v2` applications to be created for migration staging
- build their existing `_repo/apps/<slug>` source into the normal application dist store
- preserve the legacy scaffold path for `inline_v1` applications
- cover creation, serving, and Solution-capture eligibility end to end

## Why

The supported v1-to-Solution migration sequence needs a temporary loose v2 application so it can run against existing loose tables, be proven in place, participate in a slug swap, and then be captured into a Solution. The repository previously rejected this state even though swap and capture already supported it.

## Validation

- `./test.sh tests/unit/repositories/test_applications_repository.py -v` — 4 passed
- `./test.sh tests/e2e/platform/test_loose_v2_app_e2e.py -v` — 1 passed
- `./test.sh quality api` — 0 errors, all checks passed
- `git diff --check`

Docs freshness reports pre-existing CLI drift in `api/bifrost/cli.py`; this API repository change does not add or change CLI flags.

### DIFF
--- a/api/src/repositories/applications.py
+++ b/api/src/repositories/applications.py
@@ -7,6 +7,7 @@ HTTP-handler-only surface. See docs/plans/2026-05-26-org-scoping-consolidation.m
 
 from __future__ import annotations
 
+import asyncio
 import logging
 from datetime import datetime, timezone
 from uuid import UUID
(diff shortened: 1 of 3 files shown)


jackmusick:
Closing this: the migration should use a Solution-managed v2 app with global_repo_access enabled, which intentionally falls back to unresolved global _repo resources during side-by-side validation. No loose-v2 platform change is needed.

PR #501, opened 2026-07-18
fix(knowledge): make document update work and be chunk-aware end to end

### LINKED ISSUE #498: [bug]: Editing a knowledge document 500s — update path stores a nested vector and never embeds
**Summary**
Updating a knowledge document (`PUT /api/knowledge-sources/{namespace}/documents/{id}`) fails with a 500; the edit is never saved and the document is left un-re-embedded.

**Steps to reproduce**
1. Configure an embedding provider (System Settings → AI Configuration).
2. Create a knowledge document in a namespace (this succeeds).
3. Edit that document — via the UI, or `PUT /api/knowledge-sources/{namespace}/documents/{id}` with a new `content`.

**Expected behavior**
The document is re-embedded and saved (same as create), returning 200.

**Actual behavior**
500 Internal Server Error. The request rolls back, so the edit is lost. Server log:

```
sqlalchemy.exc.StatementError: (builtins.TypeError) float() argument must be a string or a real number, not 'list'
  File "src/routers/knowledge_sources.py", line 500, in update_document
  File "src/models/orm/vector_type.py", line 26, in <genexpr>
[SQL: UPDATE knowledge_store SET content=$1, embedding=$2, updated_at=$3 WHERE knowledge_store.id = $4]
```

**Root cause**
In `update_document` (`api/src/routers/knowledge_sources.py:500`):

```python
embedding = await client.embed(data.content)   # embed() returns list[list[float]]
...
doc.embedding = embedding                       # assigns the whole batch to one row
```

`BaseEmbeddingClient.embed()` takes a `list[str]` and returns a `list[list[float]]` (one vector per input). The result is assigned directly to a single row's `embedding` column, so the `Vector` bind processor calls `float()` on an inner list and raises. (The `create_document` path avoids this by going through `KnowledgeRepository.store_chunked`.) The update path also never chunks, so even without the crash it would collapse long content into a single row — diverging from create.

**Environment**
`main` @ 37bd1f71, dev stack (`./debug.sh`), embedding provider configured. Reproduced against the running API.

---

*(I have a fix ready with regression tests — happy to open a PR if you'd like.)*

pszatkowski:
Fixes #498

## Summary

`PUT /api/knowledge-sources/{ns}/documents/{id}` crashed on every call: it embedded the content as a single batch and assigned the resulting `list[list[float]]` to one row's vector column (`TypeError` at flush), and unlike create it never chunked. Fixing update properly surfaced a broader gap: **documents have been multi-row (chunked) entities since chunking landed, but the read, delete, and conflict-check paths still assumed one row per document.** This PR makes the whole document surface chunk-aware.

## What was broken on `main`

- **Update crashed unconditionally** — the endpoint has never successfully updated a document since chunking was introduced (#498).
- **Reads truncated or crashed on chunked docs**: `GET /documents/{id}` returned only the addressed chunk row's content (~2000 chars of a long document); the SDK get-by-key raised `MultipleResultsFound`.
- **DELETE removed only the addressed row** of a multi-chunk document, orphaning sibling chunks that kept surfacing in search.
- **Scope-move conflict pre-check** matched every chunk row of a conflicting doc (`MultipleResultsFound` → 500), never checked keyless collisions (NULL keys are equal under the `NULLS NOT DISTINCT` constraint, so they conflict too), and `replace=true` deleted only one row of the occupant.
- **Create's response echoed the first chunk's slice** instead of the submitted content.

## What this PR does

**Update goes through `KnowledgeRepository.replace_chunked`** — the same chunk → embed → store path as create. The document's rows are replaced within the request transaction while keeping its public identity: the first new row reuses the document id, `created_at` is preserved (edits don't reorder listings), and a scope change moves the whole document. The embed runs before the delete and everything is flush-only, so a failed embed leaves the document untouched (embed-time `ValueError` keeps the 503 contract). Concurrent edits of the same document serialize on `FOR UPDATE`, since delete-then-insert would otherwise race on the unique constraint. Keyed and keyless documents behave identically.

**Reads reassemble full content.** `split_into_chunks` gained an exact inverse, `reassemble_chunks` — provable, not heuristic: `_find_boundary` floors every non-final chunk at half the window, so consecutive chunks always overlap by exactly `overlap_chars` (now guarded at split time). No schema migration; the read-config-must-match-store-config constraint is documented, mirroring the existing `embedding`-column precedent. `get_by_id` and `get_by_key` now return the original document.

**Conflict handling and DELETE are document-scoped**: the pre-check resolves the conflicting document via its `chunk_index == 0` row and handles keyless identities (409 with the standard payload instead of a raw constraint violation); `replace=true` and DELETE remove every chunk row.

**Data access lives in the repository**: document mutation/lookup queries move from inline router SQL into `KnowledgeRepository` (`replace_chunked` / `find_document_id` / `delete_document`), per the repository pattern.

**Search is untouched**: `repo.search()` and every workflow/agent/MCP search entry point are byte-identical to main; rows written by the new update path have the same shape search already consumes.

## Testing

- 15 unit tests for the update flow against a real DB (multi-chunk GET→PUT round-trip, id stability, keyless multi-chunk replace, keyed + keyless scope-move conflicts with and without `replace=true`, embed-failure rollback, 503 mapping), plus round-trip property tests for `reassemble_chunks` including adversarial repetitive text.
- 2 new e2e tests for the PUT surface, gated on `EMBEDDINGS_AI_TEST_KEY` like the rest of the knowledge e2e module.
- Full backend unit (5321 passed) and e2e (1684 passed) suites green; pyright/ruff/tsc/lint clean; `v1.d.ts` regenerated (JSDoc-only diff).

## Out of scope (pre-existing, follow-up candidates)

- The bulk scope-update endpoint moves rows by id and can split a multi-chunk document's chunks across orgs.
- `list_documents` shows each chunk row of a multi-chunk document as a separate entry.

### DIFF
--- a/api/src/repositories/knowledge.py
+++ b/api/src/repositories/knowledge.py
@@ -15,7 +15,7 @@ from sqlalchemy import delete, func, select
 from src.models.orm import KnowledgeStore
 from src.repositories.org_scoped import OrgScopedRepository
 from src.services.embeddings import BaseEmbeddingClient
-from src.services.knowledge.chunking import split_into_chunks
+from src.services.knowledge.chunking import reassemble_chunks, split_into_chunks
 
 
 @dataclass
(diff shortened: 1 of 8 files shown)


jackmusick (approved):
Paul, thank you — this is an exceptionally thorough first contribution. The chunk-aware replacement path, full-content reassembly, document-scoped delete/conflict handling, stable identity work, and regression coverage go well beyond the original report. I ran the focused repository/router/chunking tests and API quality checks locally; they are green.\n\nI'm approving and merging this as the primary fix for #498 so the contribution and credit stay here. During adversarial review we found two narrow hardening cases that maintainers will land immediately afterward: always resolve/lock and preserve the canonical chunk-0 ID when a request arrives through a sibling chunk ID, and preserve the original created_by value when recreating rows. We'll own those follow-ups rather than asking you to expand an already substantial PR.\n\nReally appreciate the quality of the report, implementation, and tests — welcome contribution.

PR #536, opened 2026-08-01
fix(llm): stream inside AnthropicClient.complete to avoid the non-streaming 10-minute guard

sdc53:
## Problem

`AnthropicClient.complete()` calls `self.client.messages.create(**kwargs)` (non-streaming). The Anthropic SDK refuses a non-streaming request whenever `max_tokens` is large enough to risk a >10-minute completion, raising:

> Streaming is required for operations that may take longer than 10 minutes

That guard fires once `max_tokens` exceeds ~21k (`600s ÷ 3600s × 128000`).

`complete()` is used by **run summarization** (`services/execution/run_summarizer.py`), which passes no `max_tokens` and therefore inherits the admin-configured `LLMConfig.max_tokens`. When an admin sets that above ~21k (e.g. 32768 for a 128k-output model), **every run-summary generation fails** with the guard error — persisted as `summary_error = "LLM call failed: Streaming is required…"` on the run. The default (16384) is under the ceiling, so this only bites deployments that raised the limit.

## Fix

Switch `complete()` to `messages.stream()` + `get_final_message()`. The caller still gets one fully assembled `LLMResponse` (same return contract — text, tool calls, usage, stop reason, model), but the streaming path has no 10-minute ceiling, so it holds at any `max_tokens` **without capping output**. This is Anthropic's own recommended pattern for large-`max_tokens` requests.

## Tests

Two new unit tests in `test_anthropic_client_messages.py`:
- `test_complete_uses_streaming` — asserts `complete()` calls `messages.stream` and **never** `messages.create`, forwards `max_tokens` verbatim, and preserves the return contract.
- `test_complete_assembles_tool_calls` — asserts tool-use blocks are still assembled from the final message.

All anthropic-client unit tests pass; `ruff` and `pyright` clean on the changed files.

https://claude.ai/code/session_013673rJgafX4A7TCnv1QbSa

### DIFF
--- a/api/src/services/llm/anthropic_client.py
+++ b/api/src/services/llm/anthropic_client.py
@@ -51,7 +51,18 @@ class AnthropicClient(BaseLLMClient):
         max_tokens: int | None = None,
         model: str | None = None,
     ) -> LLMResponse:
-        """Non-streaming completion via Anthropic API."""
+        """Non-streaming completion via Anthropic API.
+
+        Uses ``messages.stream()`` + ``get_final_message()`` under the hood
+        rather than ``messages.create()``: the caller still receives one fully
+        assembled ``LLMResponse``, but streaming avoids the Anthropic SDK's
+        client-side guard that refuses a *non-streaming* request whenever
+        ``max_tokens`` is large enough to risk a >10-minute completion
+        ("Streaming is required for operations that may take longer than 10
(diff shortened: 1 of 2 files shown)


jackmusick (approved):
Reviewed locally: reproduced the Anthropic SDK guard, verified the real SDK streaming path preserves the LLMResponse contract at high max_tokens, and ran the targeted Anthropic client and summarizer suites. No blocking findings.

PR #537, opened 2026-08-01
fix(workflows): honor a configured 0 for cache_ttl_seconds and timeout_seconds

sdc53:
## Summary

`_convert_workflow_orm_to_schema` coalesces the data-provider cache TTL with `or 300`, so a provider explicitly configured **not** to cache is reported by the API as caching for 5 minutes.

`0` is a documented, accepted value here rather than "unset":

- `PATCH /api/workflows/{id}` documents `cache_ttl_seconds` as `0-86400 seconds` and validates that range (`api/src/routers/workflows.py`), so `0` is stored as `0`.
- The execution engine gates on `if is_data_provider and request.cache_ttl_seconds > 0` (`api/src/services/execution/engine.py`), so a stored `0` genuinely disables caching.

The read path contradicted both and reported `300`.

## Why it's worth fixing beyond the cosmetics

`cache_ttl_seconds` round-trips through the workflow settings UI. Opening a deliberately-uncached provider and saving the form writes the displayed `300` back to the database, silently re-enabling caching on a provider that had been opted out — surfacing later as stale data with no visible cause.

## Prior art in this repo

This is the same defect #27 fixed one line above for `timeout_seconds` (where `0 = no timeout`):

```diff
-        timeout_seconds=workflow.timeout_seconds or 1800,
+        timeout_seconds=workflow.timeout_seconds if workflow.timeout_seconds is not None else 1800,
```

`api/bifrost/manifest.py` already guards the same trap explicitly:

```python
# NOT `or 1800` — 0 means "no timeout" and `or` would clobber it.
timeout_seconds=wf.timeout_seconds if wf.timeout_seconds is not None else 1800,
```

The adjacent cache TTL was missed in both passes. This brings it in line and adds the comment in the matching style.

## Testing

New `api/tests/unit/test_workflow_metadata_ttl.py` covers `0` / `None` / explicit for `cache_ttl_seconds`, plus a regression guard for the `timeout_seconds` case from #27.

Verified non-vacuous: with the source fix reverted and the tests kept, `test_zero_ttl_is_preserved_not_defaulted` fails with `AssertionError: assert 300 == 0`.

`ruff check` and `pyright` are clean on both changed files.

### DIFF
--- a/api/src/routers/endpoints.py
+++ b/api/src/routers/endpoints.py
@@ -155,7 +155,11 @@ async def execute_endpoint(
                 workflow_id=str(workflow.id),
                 file_path=workflow.path,
                 execution_mode=workflow.execution_mode or "sync",
-                timeout_seconds=workflow.timeout_seconds or 1800,
+                # NOT `or 1800` — 0 means "no timeout", and `or` would clobber
+                # it *and* persist the wrong value into the Redis cache below.
+                timeout_seconds=(
+                    workflow.timeout_seconds if workflow.timeout_seconds is not None else 1800
+                ),
                 allowed_methods=workflow.allowed_methods or ["POST"],
             )
 
--- a/api/src/routers/workflows.py
+++ b/api/src/routers/workflows.py
@@ -128,7 +128,10 @@ def _convert_workflow_orm_to_schema(workflow: WorkflowORM, used_by_count: int = 
         public_endpoint=workflow.public_endpoint or False,
         is_tool=workflow.type == "tool",  # Derive from type field
         tool_description=workflow.tool_description,
-        cache_ttl_seconds=workflow.cache_ttl_seconds or 300,
+        # NOT `or 300` — 0 means "never cache" and `or` would clobber it.
+        cache_ttl_seconds=(
+            workflow.cache_ttl_seconds if workflow.cache_ttl_seconds is not None else 300
+        ),
         time_saved=workflow.time_saved or 0,
         value=float(workflow.value or 0.0),
         used_by_count=used_by_count,
--- a/api/src/services/workflow_validation.py
+++ b/api/src/services/workflow_validation.py
@@ -79,7 +79,12 @@ def _convert_workflow_metadata_to_model(
         tags=workflow_metadata.tags or [],
         parameters=parameters,
         execution_mode=workflow_metadata.execution_mode,
-        timeout_seconds=workflow_metadata.timeout_seconds or 1800,
+        # NOT `or 1800` — 0 means "no timeout" and `or` would clobber it.
+        timeout_seconds=(
+            workflow_metadata.timeout_seconds
+            if workflow_metadata.timeout_seconds is not None
+            else 1800
+        ),
         retry_policy=None,
         schedule=None,
         endpoint_enabled=False,
--- a/api/tests/unit/test_timeout_zero_not_defaulted.py
+++ b/api/tests/unit/test_timeout_zero_not_defaulted.py
@@ -1,0 +1,85 @@
+"""`timeout_seconds = 0` means "no timeout" and must survive every read path.
+
+The execution stack implements 0-as-no-timeout deliberately:
+
+- ``services/execution/service.py`` preserves it with an ``is not None`` check
+  in three separate places.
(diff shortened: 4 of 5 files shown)


sdc53:
Confirming the UI round-trip concretely, since it's the part that turns this from cosmetic into a data-freshness bug.

`client/src/components/workflows/WorkflowEditDialog.tsx` exposes the field per data provider, in a "Data Provider Config" tab rendered only when `isDataProviderType`:

```tsx
<Label htmlFor="cache-ttl">Cache TTL (seconds)</Label>
<Input id="cache-ttl" type="number" min={0} max={86400}
       value={cacheTtlSeconds}
       onChange={(e) => setCacheTtlSeconds(Number(e.target.value))} />
<p className="text-xs text-muted-foreground">
  How long to cache results (0-86400 seconds, default 300). Set to 0 to disable caching.
</p>
```

So the UI itself documents **"Set to 0 to disable caching"**, with `min={0}`.

Notably the client already handles 0 correctly — it populates with nullish coalescing, which preserves a stored 0:

```tsx
setCacheTtlSeconds(workflow.cache_ttl_seconds ?? 300);   // line 217
```

and sends it back on save:

```tsx
cache_ttl_seconds: cacheTtlSeconds,                      // line 272
```

The server's `or 300` was defeating that deliberate `??`. The resulting sequence:

1. Set 0 in the dialog → stored as 0 → caching correctly disabled at execution.
2. Reopen the dialog → `GET` reports 300 → the field displays 300.
3. Save the dialog again for any unrelated reason (rename, tags, access level) → 300 is written back → caching silently re-enabled on a provider that was opted out.

Step 3 needs no interaction with the cache field at all, which is why this surfaces later as unexplained stale data. No client change is needed — the fix is entirely the read path in this PR.

sdc53:
Expanded scope — I went looking for the same defect elsewhere and found it on `timeout_seconds`, the field #27 fixed. Retitled accordingly. Two more read paths, and a guard so the class can't come back.

### `routers/endpoints.py` — the damaging one

```python
workflow_metadata = CachedWorkflowMetadata(
    ...
    timeout_seconds=workflow.timeout_seconds or 1800,   # <-- 0 becomes 1800
)
# ...then persisted:
await redis_client.set_endpoint_workflow_cache(... timeout_seconds=workflow_metadata.timeout_seconds ...)
```

A workflow configured for no timeout is capped at 30 minutes when invoked through its HTTP endpoint. The coerced value is then written into the Redis endpoint cache, so the fast path at the top of `execute_endpoint` (which reads `cached["timeout_seconds"]` verbatim, correctly) faithfully replays the wrong value on every subsequent request.

`services/workflow_validation.py` had the same coalescing on the pending-workflow validation response — display only, included for consistency.

### That `0` is a real contract, not a quirk

| Site | Code |
|---|---|
| `services/execution/service.py:103,116,197` | `timeout_seconds if ... is not None else 1800` (three times) |
| `services/execution/service.py:472` | `if workflow_meta.timeout_seconds > 0:` |
| `services/execution/process_pool.py:168,810` | `timeout_seconds > 0 and elapsed > timeout_seconds` |
| `jobs/schedulers/execution_cleanup.py:84` | `# timeout_seconds == 0 means no timeout — skip entirely` |

The execution stack goes out of its way to preserve 0; these two paths were the outliers.

### Source guard

`test_no_falsy_coalescing_on_zero_meaningful_fields` scans `api/src` for `<field> ... or <nonzero>` on `timeout_seconds` / `cache_ttl_seconds`. Worth having because the endpoint metadata builder is inline in a router that needs redis + db to exercise, so it has no cheap unit-test seam — yet it was the more damaging of the two sites. The guard covers it, and any future one.

It's also self-demonstrating: written on a branch without the first commit, it failed on `routers/workflows.py:131` — the original `cache_ttl` bug. Which is why both fixes are in this one PR rather than split; the guard can't pass without both.

I found this because three workflows in our own deployment are configured `timeout_seconds=0` with `endpoint_enabled=true`, so they were silently running under a 30-minute cap.

jackmusick (approved):
Reviewed the zero-value semantics across the workflow API, endpoint metadata cache, Redis wait path, execution engine, and cleanup behavior. The changes preserve the documented meaning of 0, and targeted unit/API/endpoint E2E coverage plus API quality checks pass locally. No blocking findings.

PR #586, opened 2026-08-13
Add verified Resend webhook adapter

jackmusick:
## What changed

- adds a built-in Resend webhook adapter with Standard Webhooks/Svix signature verification
- supports callback-first provisioning and later attachment of the Resend signing secret
- adds a runtime-only webhook configuration endpoint usable by Solution-managed sources
- preserves instance-owned webhook state across idempotent Solution redeploys
- adds unit and end-to-end coverage for signature handling, callback-first configuration, and redeploy preservation

## Why

The DMARC Solution deploys its Resend event source and workflow subscription. Resend reveals its signing secret only after receiving the callback URL, so the platform needs a safe post-deploy runtime-secret path without turning the source into a loose entity. Runtime state must also survive later Solution redeploys.

## Validation

- `./test.sh tests/unit/services/test_resend_webhook_adapter.py tests/unit/test_solution_capture.py::test_redeploy_webhook_preserves_instance_state tests/e2e/api/test_events.py::TestEventSourceCRUD::test_resend_source_can_add_signing_secret_after_callback -v` — 13 passed
- `./test.sh tests/unit/test_dto_flags.py tests/unit/test_contract_version.py -v` — 64 passed
- `./test.sh quality api` — 0 errors, 0 warnings


jackmusick:
Closing at the request of the project owner; this platform change is not needed.

PR #646, opened 2026-08-25
fix(execution): preserve optional import misses

### LINKED ISSUE #645: [bug]: Resolver outage breaks optional imports in installed dependencies
### Summary

The virtual import finder turns an optional dependency miss into a hard workflow failure when workspace resolution is unavailable.

### Steps to reproduce

1. Install `grpcio` without the optional `grpcio-tools` companion.
2. Enable Bifrost's virtual import finder.
3. Make `resolve_module_sync()` raise `ModuleResolutionError`.
4. Import code that probes a missing module inside `try/except ImportError`.

### Expected behavior

The virtual finder logs the resolver failure and returns `None` for every caller, allowing Python's normal importer chain to continue. If no finder provides the module, Python raises `ModuleNotFoundError` whether the import originated in an installed dependency, standard-library module, or virtual workspace module.

### Actual behavior

`ModuleResolutionError: Module resolver unavailable for grpc_tools` escapes the import hook and aborts the execution.

### Environment

`gobifrost/bifrost` main at `7256d1db2`; behavior introduced by `dccd6f0e3`. Reproduced in a worker with `grpcio==1.83.0` and no `grpc_tools`.

### Logs / screenshots

```text
Module resolver unavailable for grpc_tools
```

MTG-Thomas:
## Summary

- preserve normal Python import behavior for every caller when the workspace resolver is unavailable
- log the requested module and original resolver exception for operators before continuing the import chain
- cover installed-package, standard-library, and virtual-workspace optional imports

## Root cause

The targeted virtual-import finder introduced in #621 consults the workspace resolver after Python's path finder misses a module. Code commonly probes optional companions inside `try`/`except ImportError`; when the resolver is unavailable, `ModuleResolutionError` currently escapes instead of allowing Python to continue through its importer chain. A `grpcio` import probing the absent `grpc_tools` package exposed the regression.

The finder now logs the resolver exception and returns `None` uniformly. Another finder may still provide the module. If none does, Python raises the normal `ModuleNotFoundError`, regardless of whether the import originated in an installed dependency, standard-library module, or virtual workspace module.

## Verification

- `python -m pytest -q tests/unit/services/test_virtual_import.py` — 38 passed
- `ruff check api/src/services/execution/virtual_import.py api/tests/unit/services/test_virtual_import.py`
- `python -m py_compile api/src/services/execution/virtual_import.py api/tests/unit/services/test_virtual_import.py`

## Review note

This touches the execution engine. The tests exercise the resolver-outage boundary for installed, standard-library, and workspace callers, and verify that the operator log retains the original exception.

Fixes #645

### DIFF
--- a/api/src/services/execution/virtual_import.py
+++ b/api/src/services/execution/virtual_import.py
@@ -352,15 +352,15 @@ class VirtualModuleFinder(MetaPathFinder):
         try:
             resolution = resolve_module_sync(fullname)
         except ModuleResolutionError:
-            # Installed dependencies commonly probe optional companions inside
-            # ``try: import ... except ImportError`` blocks. If the workspace
-            # resolver is temporarily unavailable, preserve that normal Python
-            # contract for an import originating outside workspace/platform
-            # code. Workspace imports still fail closed so a cache/API outage
-            # cannot silently masquerade as a missing workspace module.
-            if _import_origin_is_external():
-                return None
-            raise
+            # A finder that cannot provide a module must let Python continue
+            # through the normal import chain. Keep the resolver failure visible
+            # to operators without exposing platform-specific import semantics
+            # to workspace or dependency code.
+            logger.exception(
+                "Virtual module resolution failed for %s; continuing normal import",
+                fullname,
+            )
+            return None
         if resolution.kind in {"module", "package"} and resolution.content is not None:
             is_package = resolution.kind == "package"
             loader = VirtualModuleLoader(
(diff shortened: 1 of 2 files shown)


jackmusick:
Thanks for digging into this. The direction is right, but I’d like the behavior to be uniform rather than depend on where the import originated.

One of Bifrost’s platform principles is that Python should behave like Python from the developer’s perspective, with as few platform-specific exceptions as possible. If the virtual importer cannot provide a module because resolution failed, it should return `None` for every caller and allow Python’s normal import chain to continue. If no importer can provide the module, Python will naturally raise `ModuleNotFoundError`, whether the import came from a workflow, another virtual module, the standard library, or an installed package.

Please remove the caller-origin classification and update the tests to cover identical behavior for installed-package and workspace/workflow imports.

We should still preserve operational visibility: log the resolver failure internally with the requested module name and original exception (and retain whatever metrics are appropriate), while keeping the developer-facing import behavior standard. The goal is uniform Python semantics plus diagnostic evidence for operators.

MTG-Thomas:
Agreed. I removed the caller-origin classification. A `ModuleResolutionError` now logs the requested module and original exception, then returns `None` so Python continues its normal importer chain for every caller.

I also changed the virtual-workspace regression to exercise the same `try`/`except ImportError` behavior as the installed-package case and assert that the operator log retains the original exception. The focused importer suite remains green: 38 passed, with Ruff and byte-compilation clean.

Updated in `92112a144`.

Issue #652, opened 2026-08-26
[bug]: Event subscription filter_expression is persisted but never enforced


MTG-Thomas:
### Summary

Event subscriptions accept, persist, export, import, and advertise an optional JSONPath `filter_expression`, but event processing never evaluates it. A subscription with a non-matching filter still queues its workflow or agent target.

### Steps to reproduce

1. Create a webhook or topic event source.
2. Subscribe a workflow with the matching event type and `filter_expression: "$.priority == 'high'"`.
3. Emit an otherwise matching event whose payload contains `priority: "low"`.
4. Inspect the event deliveries and workflow executions.

### Expected behavior

The subscription filter is evaluated before target dispatch. A non-matching event does not queue the target. Invalid or unevaluable filters fail closed and leave an observable decision instead of silently running automation.

### Actual behavior

Both `EventProcessor.emit_topic()` and `EventProcessor._process_delivery()` select active subscriptions by source and event type, then create pending deliveries without reading `subscription.filter_expression`. `queue_event_deliveries()` consequently queues the target.

This is especially surprising because the field is exposed through the REST contracts, CLI, manifest round-trip, and Solution deployment, and is described as a JSONPath expression. Some contract text calls it "future use", but no mutation surface rejects it or warns that it is inert.

### Environment

`gobifrost/bifrost` `main` at `7fbda650359c9bdbacb0b4cc373cd4c3b3ff6ff3`.

### Logs / screenshots

No exception is emitted. Code inspection shows neither delivery-creation path evaluates `filter_expression`; a linked draft PR will add regression coverage and a proposed fail-closed implementation.

PR #654, opened 2026-08-27
docs(events): propose enforceable subscription rules

MTG-Thomas:
## Summary

- documents the current `filter_expression` contract gap: subscriptions accept and persist filters, but event routing does not evaluate them
- proposes versioned, structured subscription criteria with strict and fail-closed evaluation semantics
- defines one evaluator across webhook, topic, and scheduled ingress, with durable decision evidence for matched, non-matched, and evaluation-error outcomes
- outlines a guarded migration that refuses to silently reinterpret non-empty legacy filter strings

This is a design-only draft intended to establish the upstream contract before implementation. It incorporates lessons from comparable event and workflow systems, including our Celery and StackStorm integration work, while keeping the proposal native to Bifrost's existing event model.

## Why

Today a caller can configure a filter such as `$.priority == 'high'` and receive a successful response, yet a low-priority event still creates a delivery and can execute the target. That is a dangerous false assurance at an execution boundary.

The proposed contract makes these invariants explicit:

- non-matching events never execute a target
- malformed rules and evaluation errors fail closed
- routing decisions remain observable without turning skipped events into retryable failures
- retries and duplicate events cannot bypass the decision

## Scope and risk

This PR changes documentation only. It proposes a future change to sensitive event-routing and workflow-execution behavior, but does not alter runtime behavior, schemas, or APIs.

## Validation

- candidate commit: `0cc698985acc80b8178a722d1928e0f4ca22b712`
- `./test.sh pre-pr` passed against that exact commit on Debian 13 with Docker
- client: 290 files / 1,861 tests passed
- API unit: 5,770 passed / 3 skipped
- live integration and end-to-end: 1,739 passed / 40 skipped
- production client smoke and local API candidate image build passed

Related to #652.


jackmusick:
Closing as superseded by #653, which now enforces the existing filter_expression contract and closes #652. The structured, versioned criteria model proposed here is a broader future design; if we decide to pursue it, it should return as a separate RFC grounded in the now-enforced behavior.

PR #667, opened 2026-08-28
Use Pydantic's default OpenAI transport

### LINKED ISSUE #666: [feature]: Support OpenAI Responses API transport per model profile
## Problem

Bifrost currently routes OpenAI-compatible providers through `OpenAIChatModel` only. Some Microsoft Foundry deployments, including GPT-5-family models, require or work correctly through the OpenAI Responses API instead of Chat Completions.

## Proposed behavior

Add a model-profile transport setting with `auto`, `responses`, and `chat_completions` values. `auto` should select Responses for GPT-5 and reasoning-model families while preserving Chat Completions for legacy models. Explicit overrides should remain available for provider-specific compatibility.

For privacy, Responses requests should set `openai_store=false` by default.

## Prior implementation

We implemented, tested, and deployed this in our fork:

- Fork issue: https://github.com/MTG-Thomas/bifrost/issues/618
- Fork PR: https://github.com/MTG-Thomas/bifrost/pull/619

I will follow with an upstream-targeted PR based on current upstream `main`.

MTG-Thomas:
## Summary

- let Pydantic AI choose its default OpenAI model transport
- preserve Bifrost's injected OpenAI-compatible client and provider configuration
- disable OpenAI Responses API server-side storage across agent and direct-client paths
- remove the proposed profile transport setting, migration, API plumbing, and UI selector

This follows maintainer feedback: Bifrost should not force transport selection when Pydantic AI can negotiate the current default. With Pydantic AI 2.33.0, the `openai:` inference path selects `OpenAIResponsesModel` while retaining Bifrost's custom Foundry endpoint/client.

Closes #666.

Fork implementation and production context:
- https://github.com/MTG-Thomas/bifrost/issues/618
- https://github.com/MTG-Thomas/bifrost/pull/620

## Verification

- `29 passed` in `api/tests/unit/services/test_agent_runtime.py`
- focused Pyright and Ruff passed
- live Microsoft Foundry plain, tool, two-tool, and structured-output checks passed for DeepSeek V4 Flash 0731, GPT-5.6 Terra, and GPT-5.4 Mini
- GPT-5.6 Luna passed plain, one-tool, and structured-output checks; an 8-way concurrent two-tool burst completed 7/8 runs, with one 60-second timeout and successful runs completing in about 5-9 seconds
- every live run resolved through `OpenAIResponsesModel`

The Luna timeout appears to be an isolated concurrency/latency signal rather than a transport-selection failure: the same chain passed before and after the burst.

### DIFF
--- a/api/src/services/agent_runtime/model_factory.py
+++ b/api/src/services/agent_runtime/model_factory.py
@@ -6,7 +6,7 @@ scheduler import boundary while giving every agent surface one model abstraction
 
 from decimal import Decimal, InvalidOperation
 
-from pydantic_ai.models import Model
+from pydantic_ai.models import Model, infer_model
 from pydantic_ai.usage import RequestUsage
 
 from src.services.agent_runtime.retry_transport import get_ai_retry_http_client
@@ -34,6 +34,8 @@ def agent_model_settings(
         settings["max_tokens"] = resolved_max_tokens
     if is_openrouter_endpoint(config.endpoint):
         settings["extra_body"] = {"session_id": session_id[:256]}
+    elif config.provider == "openai":
+        settings["openai_store"] = False
     return settings
 
 
@@ -141,7 +143,6 @@ def create_agent_model(config: LLMConfig, *, model: str | None = None) -> Model:
             )
 
         from openai import AsyncOpenAI
-        from pydantic_ai.models.openai import OpenAIChatModel
         from pydantic_ai.providers.openai import OpenAIProvider
 
         client = AsyncOpenAI(
@@ -151,7 +152,10 @@ def create_agent_model(config: LLMConfig, *, model: str | None = None) -> Model:
             max_retries=0,
         )
         provider = OpenAIProvider(openai_client=client)
-        return OpenAIChatModel(model_name, provider=provider)
+        return infer_model(
+            f"openai:{model_name}",
+            provider_factory=lambda _provider_name: provider,
+        )
 
     if config.provider == "anthropic":
         from anthropic import AsyncAnthropic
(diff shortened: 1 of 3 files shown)


jackmusick:
I tested this against a real Microsoft Foundry deployment (`gpt-5.6-luna`, version `2026-07-09`) in the GoCovi Partner Benefits subscription.

Findings:
- Pydantic AI successfully normalizes both OpenAI transports behind Bifrost's existing interface.
- Luna supports both Responses and Chat Completions; Responses worked end-to-end through `PydanticAIClient`.
- Foundry's `/openai/v1/models` response is a broad catalog, not the resource's deployment aliases, so Bifrost must not offer those values as selectable deployment names.

I implemented the requested automatic behavior in commit gobifrost/bifrost@306b09e6d on branch `codex/azure-foundry-auto-transport`:
- probe Responses on first use of an OpenAI-compatible model profile;
- fall back only when Responses is definitively unsupported;
- store the successful transport per profile;
- clear it when the model, endpoint, provider, or key changes;
- keep Foundry deployment entry manual rather than showing its misleading model catalog.

Live verification confirmed detection was stored and reused from a second DB session without probing again, followed by a successful normal Bifrost completion. Focused tests: 56 passed; transport-specific tests: 4 passed; Pyright and Ruff passed; Alembic has one head and the new migration applied cleanly.

GitHub rejected my direct push to the Midtown organization fork with HTTP 403 despite `maintainerCanModify=true`, so the commit is on the upstream branch above and can be cherry-picked onto this PR.

jackmusick:
Follow-up on the proposed auto-transport changes: the model picker now explains that provider catalogs can include models the connected account cannot access. This is included in commit 3f271670d on `gobifrost:codex/azure-foundry-auto-transport`, with focused Vitest and Playwright coverage.

jackmusick:
Superseded by #694, which contains the tested automatic transport detection and model-catalog clarification on an upstream-owned branch. Closing this copy because GitHub rejected maintainer pushes to the organization-owned fork branch.

PR #669, opened 2026-08-29
Route multitenant webhooks by integration mapping

jackmusick:
## Summary
- let webhook adapters supply a signed external tenant/entity ID
- resolve that ID through the source integration mapping and stamp the event organization
- execute global workflow/agent targets in the mapped customer organization while preventing cross-org delivery
- expose the resolved organization on event API responses
- preserve fixed-tenant Microsoft Bot Framework sources

## Verification
- 47 focused unit tests passed
- integration-routed webhook E2E passed
- legacy fixed-tenant Teams auth E2E passed
- 64 DTO/contract tests passed
- API pyright + ruff passed
- client TypeScript passed
- generated OpenAPI client types updated

### DIFF
--- a/api/src/models/contracts/events.py
+++ b/api/src/models/contracts/events.py
@@ -451,6 +451,10 @@ class EventResponse(BaseModel):
 
     id: UUID = Field(..., description="Event ID")
     event_source_id: UUID = Field(..., description="Event source ID")
+    organization_id: UUID | None = Field(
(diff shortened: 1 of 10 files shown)


jackmusick:
Closing because tenant routing belongs in the Bifrost workspace workflow/integration layer for this implementation; no platform change is required.

PR #727, opened 2026-09-11
Support base64 webhook HMAC signatures

### LINKED ISSUE #726: [feature]: Accept base64-encoded webhook HMAC signatures
### Problem

Incoming webhook HMAC verification only accepts hex-encoded signatures. Senders that base64-encode the same HMAC-SHA256 digest are rejected as invalid, even when the shared secret is correct. Base64 is a common convention across PSA/RMM and cloud event vendors.

Related: HaloPSA sends a space between the prefix and the digest (sha256= <sig>), which also fails verification because the prefix is stripped by length without trimming.

The result is that users with a correctly configured secret see signature failures they can’t diagnose or work around.

### Proposed solution

Accept either canonical encoding of the same HMAC-SHA256 digest — hex or base64 — and tolerate whitespace around the prefix and header value. Verification stays constant-time, and existing hex behaviour is unchanged.

### Acceptance criteria

- [ ] Base64 signatures verify successfully, with or without a configured prefix
- [ ] HaloPSA’s sha256= <sig> form (space after the prefix) verifies successfully
- [ ] Tampered or malformed signatures are still rejected
- [ ] Existing hex signature behaviour is unchanged
- [ ] The public SDK webhook helper behaves identically to the runtime verifier
- [ ] Unit and endpoint-level test coverage for the new formats

### Design doc link (optional)

_No response_

pszatkowski:
## Summary

Webhook HMAC verification previously computed `hexdigest()` and compared it to the incoming header, so only hex-encoded signatures could ever match. Senders that base64-encode the same HMAC-SHA256 digest were rejected as invalid even with a correctly configured shared secret.

This widens the accepted input to either canonical encoding of the same digest, and tolerates the whitespace HaloPSA emits after the `sha256=` prefix.

Fixes #726

## Changes

- **`api/src/services/webhooks/protocol.py`** — `WebhookAdapter.verify_hmac_signature` now computes the raw digest once and accepts either `digest.hex()` or `base64.b64encode(digest)`. The header value is stripped, and stripped again after the configured prefix is removed.
- **`api/bifrost/webhooks.py`** — the public SDK helper mirrors the same hex-or-base64 acceptance and whitespace handling.
- **`api/src/services/webhooks/adapters/generic.py`** — configuration help text no longer claims the header must contain a hex digest.

## Design notes

- The incoming signature is compared against **canonically encoded expected values**; attacker-supplied input is never decoded. This avoids accepting alternate encodings of the same bytes (unpadded, base64url, whitespace-injected) as equivalent.
- Only standard **padded** base64 of the raw digest is accepted — not base64url, unpadded base64, or base64-of-hex. Those can be added if a real provider needs them.
- Whitespace is tolerated only around the whole value and immediately after the prefix. Whitespace *within* the digest is still rejected.
- Hex remains case-insensitive in the runtime verifier and case-sensitive in the SDK helper, preserving each one's prior behaviour. Base64 is case-sensitive in both.
- Both comparisons use `hmac.compare_digest`, so verification stays constant-time. Two comparisons run unconditionally rather than short-circuiting on encoding shape.
- The SDK helper continues to *require* the prefix when one is configured (its prior behaviour); the runtime verifier continues to strip the prefix only when present (its prior behaviour). Neither is changed by this PR.

No migration, DTO, generated TypeScript, or configuration change is required — this is purely a widening of accepted input, so existing hex senders are unaffected.

## Tests

`api/tests/unit/services/test_webhook_adapters.py`
- `test_base64_signature_with_whitespace_after_prefix`
- `test_whitespace_within_base64_signature_is_rejected`
- `test_sdk_helper_accepts_base64_with_whitespace_after_prefix`
- `test_base64_signature_with_whitespace_accepted`

`api/tests/e2e/api/test_events.py`
- `test_webhook_with_spaced_base64_hmac_accepted` — endpoint-level happy path for Halo's `sha256= <base64>` form

Existing hex coverage is unchanged and still passing.

### DIFF
--- a/api/bifrost/webhooks.py
+++ b/api/bifrost/webhooks.py
@@ -34,16 +34,15 @@ Usage in workspace/adapters/my_adapter.py:
             ...
 """
 
+import base64
+import hashlib
+import hmac
+import json
+import secrets
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from typing import Any, Callable, TypeVar
-
-import hashlib
-import hmac
-import secrets
-import json
-
 
 F = TypeVar("F", bound=type)
 
@@ -304,7 +303,7 @@ class WebhookAdapter(ABC):
         prefix: str = "sha256=",
     ) -> bool:
         """
-        Verify HMAC-SHA256 signature.
+        Verify a hex- or base64-encoded HMAC-SHA256 signature.
 
         Common pattern for webhook validation (GitHub, Stripe, etc.)
 
@@ -317,16 +316,25 @@ class WebhookAdapter(ABC):
         Returns:
             True if signature is valid
         """
-        expected = hmac.new(
+        signature = signature.strip()
+        if prefix:
+            if not signature.startswith(prefix):
+                return False
+            signature = signature[len(prefix) :].strip()
+
+        digest = hmac.new(
             secret.encode(),
             payload,
             hashlib.sha256,
-        ).hexdigest()
+        ).digest()
 
-        if prefix:
-            expected = f"{prefix}{expected}"
+        hex_matches = hmac.compare_digest(signature, digest.hex())
+        base64_matches = hmac.compare_digest(
+            signature,
+            base64.b64encode(digest).decode("ascii"),
+        )
 
-        return hmac.compare_digest(expected, signature)
+        return hex_matches or base64_matches
 
     @staticmethod
     def expiration_datetime(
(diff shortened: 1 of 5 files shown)


jackmusick (approved):
The HaloPSA/base64 HMAC path is correctly scoped and implemented. The verifier accepts canonical hex and padded base64 forms, tolerates the provider's whitespace after `sha256=`, preserves existing behavior, and keeps constant-time comparisons. Focused unit and endpoint verification passed locally, and CI/CodeQL are green. The parallel authoring facade is not an active custom-adapter execution surface today, so its pre-existing semantic differences are not a blocker for this fix.

PR #744, opened 2026-09-13
fix(events): give delivery retries a fresh timeout window

MTG-Thomas:
Retrying an old failed delivery resets its status but leaves timeout detection measuring from the original creation time. The cleanup scheduler can therefore immediately treat the new attempt as stuck.

Record the current attempt start, clear previous completion state, and measure active-delivery timeouts from the attempt timestamp with creation time retained for legacy rows. Keep retry lifecycle logic in the event processor and preserve upstream error formatting. The migration adds a nullable timestamp and a partial index and follows the current upstream migration head.

Includes unit coverage for successful and failed queueing plus a database regression covering a fresh retry, an expired attempt, and a legacy row. Ported from Midtown-Technology-Group/bifrost commit 07c5a7fa9; unrelated chat-test changes and deleted legacy test suites were excluded.

Sensitive paths: database migration and event delivery lifecycle. No API response contract change.

Validation on pve-t340 VM 101: delivery retry unit/database regressions and `./test.sh pre-pr` for the exact PR head.

### DIFF
--- a/api/alembic/versions/20260825_event_delivery_attempt_started.py
+++ b/api/alembic/versions/20260825_event_delivery_attempt_started.py
@@ -1,0 +1,36 @@
+"""Track the active attempt age for retried event deliveries.
+
+Revision ID: 20260825_delivery_attempt
+Revises: 20260912_document_prefix_index
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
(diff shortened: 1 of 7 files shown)


MTG-Thomas:
Superseded by #745, which preserves the original retry-timeout fix and adds the maintainer review repairs. Closing this duplicate in favor of #745.

PR #746, opened 2026-09-14
fix(sdk): retry transient transport errors; add retry_transient opt-in

sdc53:
## Why

The SDK client (`api/bifrost/client.py`) retries transient 5xx responses on idempotent HTTP methods, but a *raised* transport error (`httpx.ReadError`, `ConnectError`, `ReadTimeout`, `PoolTimeout`, `RemoteProtocolError`) never returns a response, so it bypasses the retry loop entirely. One reset worker-to-API connection fails the whole workflow. Rolling API restarts trigger this for every in-flight request.

Idempotency is also gated by verb, but many POST/PATCH calls workflows make are idempotent in effect (an ON CONFLICT upsert, a cursor write that sets an absolute value). Those are the calls long-running sync workflows make in bulk.

## What

- Transient transport exceptions are retried with the existing backoff under the same gate as 5xx (idempotent methods).
- New opt-in kwarg `retry_transient=True` on `get/post/put/patch/delete/request` and the sync variants. The caller asserts the operation is idempotent in effect; the client then retries transport errors and transient 5xx for it. Without the flag, behaviour is unchanged.
- The kwarg is popped before the request reaches httpx.

## Evidence

On our production instance a 15-minute sync workflow hit a transport error on its final cursor PATCH about three times per run; every one cleared on the first 0.5 s retry once the opt-in was in place.

## Tests

`api/tests/unit/test_bifrost_client_retry.py`: transport error on GET retries to success; POST propagates on the first error; POST with `retry_transient=True` retries; existing 5xx cases unchanged (114 SDK unit tests pass). `pyright` and `ruff` clean on the touched files.

## Sensitive paths

Touches the SDK client (worker to API). No auth or tenancy change; retry only widens to transport errors and opt-in writes.

Note: I could not run the full `./test.sh pre-pr` gate on this machine; targeted unit tests, pyright and ruff were run for the touched files.

https://claude.ai/code/session_019KyJSi4Sj3HP4HLeUWKrFo


jackmusick (approved):
The retry policy preserves the important safety boundary: idempotent HTTP methods retry automatically, while POST/PATCH require an explicit idempotency assertion. CI and the focused local retry suite are green. Sync transport-error mirror coverage can follow separately.

PR #747, opened 2026-09-14
perf(llm): enable Anthropic prompt caching on agent and client requests

sdc53:
## Why

Agent rounds re-send the same system prompt and tool definitions on every request, and multi-turn chats re-send the growing history. The Pydantic AI path currently sends no `cache_control`, so Anthropic bills and prefills the full prompt every time; `ai_usage.cache_read_tokens` is always 0. A Tech Support style agent with a 20 KB system prompt and ~35 tools runs ~45k input tokens per round.

## What

- `agent_runtime/model_factory.py`: `anthropic_prompt_cache_settings()` returns `anthropic_cache=True` (automatic moving breakpoint on the latest message), `anthropic_cache_instructions=True`, `anthropic_cache_tool_definitions=True`. `agent_model_settings()` merges it for `provider == "anthropic"` (OpenRouter keeps its own sticky-cache settings).
- `llm/pydantic_client.py`: `_model_settings()` applies the same for `complete`/`stream`. Return type widened to `ModelSettings` since the dict is no longer OpenAI-only.
- Other providers are untouched.

## Measured

On a ~45k-token agent request, the second call read 44,514 of 44,952 input tokens from cache (cache write 434). Anthropic prices cache reads at 10% of input.

## Tests

`api/tests/unit/test_anthropic_prompt_cache.py` (new) and the updated expectation in `tests/unit/services/test_agent_runtime.py::test_runtime_uses_provider_output_defaults_except_when_api_requires_limit`. `pyright` and `ruff` clean on touched files.

Note: I could not run the full `./test.sh pre-pr` gate on this machine; targeted unit tests, pyright and ruff were run for the touched files.

https://claude.ai/code/session_019KyJSi4Sj3HP4HLeUWKrFo


jackmusick (approved):
The caching change is valuable and the direct Anthropic path is well covered by CI and focused local tests. I am comfortable shipping this now, with endpoint capability detection and a no-cache fallback handled in the immediate follow-up.

PR #755, opened 2026-09-14
fix(jobs): restore audit actor in platform runner

MTG-Thomas:
Platform-job handlers run detached from the HTTP request scope, so the audit actor context is empty and audit events emitted by durable jobs lose the requesting user's identity.

This sets the audit actor from the job's requested_by fields before invoking the handler and clears it afterwards, tagging job audit events with source platform_job.

Upstream port of Midtown-Technology-Group/bifrost@9d4ea1202 (regression test adapted: fork-only execution_policy/WorkloadClass deps removed).

Verification: api/tests/unit/jobs/test_platform_job_runner.py passes 1/1.


jackmusick (approved):
Restores the platform-job requester into the audit context with token-based cleanup in the finally block, so the prior context is reliably restored on every handler outcome. The focused regression test passed locally (1/1), and the current CI checks are green.

PR #756, opened 2026-09-14
fix(chat): preserve new conversation activity

MTG-Thomas:
Route synchronization can select a just-created conversation after its first message has already started streaming. The null to id transition is not a genuine user switch, so it must not clear the conversation-scoped streaming state.

Upstream port of the chatStore hunk from Midtown-Technology-Group/bifrost@083f4a170 (store guard plus one regression test only; ChatWindow and e2e hunks excluded).

Verification: client vitest src/stores/chatStore.test.ts passes 3/3.


jackmusick (approved):
The null-to-ID transition for a newly created conversation now preserves the already-started stream, while the existing test still verifies that a genuine conversation-to-conversation switch clears streaming state. The focused store tests passed locally (3/3), and the current CI checks are green.

Issue #685, opened 2026-09-02
[bug]: Preserve authenticated caller scope for global agent workflow tools


jackmusick:
## Summary

Follow up to #630/#631. Chat currently supplies the agent's organization separately from caller identity. Consequently, a global agent invoking a global workflow for an authenticated organization user can execute in global scope instead of the user's organization.

## Expected behavior

- An org-scoped workflow executes in the workflow's org.
- A global workflow preserves the invocation org.
- Interactive calls use the authorized explicit/user org.
- Autonomous calls without a user use the agent's org.
- A global autonomous agent invoking a global workflow remains global.
- Platform-org impersonation remains enforced by the canonical execution engine.

## Proposed approach

- Make organization scope part of the shared typed caller context introduced in #631 so identity and scope cannot be constructed independently.
- Use that shared context consistently across chat, autonomous execution, and MCP/gateway paths.
- Cover workflow-backed tools and system workflow tools.
- Verify tool enumeration remains consistent with execution authorization.
- Preserve the workflow worker's existing final workflow-org override and all public streaming/result contracts.
- Treat making chat consume the generic MCP gateway as a separate architectural follow-up.

## Acceptance criteria

- A global agent invoking a global workflow during interactive chat executes in the authenticated caller's authorized org.
- An org-scoped workflow continues to execute in its own org.
- An autonomous run without a user uses the agent's org, including when invoking a global workflow.
- A global autonomous agent invoking a global workflow remains global.
- Platform-org override/impersonation rules are not duplicated or weakened outside the canonical execution scope logic.
- Agent tool enumeration and execution agree for the affected caller roles.
- Focused unit and end-to-end regression tests cover the changed boundaries.

PR #757, opened 2026-09-14
fix(workers): drain admitted workflows before shutdown

MTG-Thomas:
Broker drain() only waits for routing handlers, then stop() tears down the pool while admitted child executions and result callbacks are still running. This adds a ProcessPoolManager.drain() that waits for admitted executions plus result persistence, a _drain_admitted_work hook in the broker consumer drain path, and the workflow consumer override.

Upstream port of Midtown-Technology-Group/bifrost@ff017a556 / #685 (code plus test only; docs excluded). The upstream worker/app.py _drain_consumer shutdown path already exists.

Verification: api/tests/unit/jobs/test_workflow_pool_drain.py ported; full pytest run blocked on host env (no pytest module), syntax checks pass.


jackmusick (approved):
Reviewed the workflow pool drain path and its focused coverage. The implementation correctly tracks active processes, waits for cooperative completion, and terminates remaining work after the drain deadline. Focused tests passed (5/5), and the existing RabbitMQ drain suite also passed (8/8). No actionable findings.

PR #758, opened 2026-09-14
fix(chat): keep expanding run activity from scrolling tool controls

MTG-Thomas:
Expanding run activity uses overflow-hidden, so focus and scrollIntoView inside the collapsing grid scroll the page and can move a tool control under the pointer. overflow-clip keeps the clipping without creating a scroll container.

Upstream port of the client hunk from Midtown-Technology-Group/bifrost@2def0ae3f (Azure hunks excluded).

Verification: ChatRunActivity.test.tsx 5/5 locally.


jackmusick (changes requested):
The cited ChatRunActivity suite passes, but it never asserts the regression contract introduced here: that the clipping wrapper uses overflow-clip rather than overflow-hidden. Because the bug is caused by that exact CSS distinction, please add a focused assertion for the wrapper class, including the absence of overflow-hidden, so a future refactor cannot silently reintroduce it.

MTG-Thomas:
Added the regression assertion: the expand test now checks the clipping wrapper has overflow-clip and not overflow-hidden. Verified 5/5 locally.

jackmusick (approved):
The focused assertion now locks down the exact overflow behavior that caused the regression, including the absence of `overflow-hidden`. My earlier concern is resolved. Looks good.

PR #759, opened 2026-09-14
fix(mcp): read scopes from nested authorization-server metadata

MTG-Thomas:
RFC8414-style discovery nests scopes_supported under authorization_server_metadata. readMetadata only checked the top level, leaving the Scopes field blank for those servers. This prefers the nested value in the fallback chain (purely additive, existing order preserved).

Upstream port of the client hunks from Midtown-Technology-Group/bifrost@1fbb02904. Note: nested endpoint lookup (issuer, token_endpoint) belongs to a separate fork change and stays out of scope; the ported test uses top-level endpoints in its mock.

Verification: MCPServerForm.test.tsx 4/4 locally.


jackmusick (changes requested):
The new test mocks authorization_server_metadata.scopes_supported, but the current /api/mcp-servers/discover backend cannot produce that response shape. discover_oauth_metadata() explicitly merges the authorization-server and protected-resource documents into one top-level dictionary, where scopes_supported is already top-level. This therefore adds a speculative fallback rather than covering a reachable regression. Please either port the backend contract that actually emits the nested shape and test that boundary, or drop this change.

MTG-Thomas:
You're right — upstream discover_oauth_metadata() merges both documents into one top-level dict, so the nested shape this fallback handles can't occur through the real backend. The fork's nested shape comes from its own discovery contract change (d08c72e07), which is a bigger port needing backend plus API review. Closing this rather than expanding scope; the nested-scopes handling can ride along if that contract is ever ported.

PR #760, opened 2026-09-14
fix(security): resolve websocket table claims per subscription

MTG-Thomas:
Websocket principals live across tables and policy changes, but custom claims can be solution-scoped and mutable. Each table evaluation must resolve claims afresh rather than sharing a name-keyed cache across subscriptions.

Manual upstream port of Midtown-Technology-Group/bifrost@99d0cea78 (adapted to upstream's diverged websocket.py; audit denial emitters untouched). Regression coverage added to the existing test_websocket_table_resolution.py.

Verification: tests/unit/test_websocket_table_resolution.py 5/5 locally.


jackmusick (changes requested):
Please add a regression test that exercises the real policy loader and demonstrates that a DB-backed custom claim controls WebSocket subscription or message visibility, including fresh resolution after the claim source changes. The current test verifies fresh principal copies but mocks away claim resolution and authorization, so it cannot catch a regression in the security fix itself.

MTG-Thomas:
Added TestClaimGatedSubscriptionVisibility: seeds a source table, a list claim scoped to the caller, and a claim-referencing policy table. A matching document_change is delivered; after the claim source changes, the same message is suppressed. Real loader, preresolution, and authorization — no mocks on that path. Verified it fails without the fix (claims never resolve, 0 deliveries) and passes with it (6/6 locally).

jackmusick (approved):
The added DB-backed regression test satisfies my earlier request: it exercises the real claim loader and authorization path and verifies fresh resolution after the source data changes. The implementation also correctly isolates resolved claims from the long-lived WebSocket principal. Looks good.

PR #761, opened 2026-09-14
fix(security): verify database TLS peers

MTG-Thomas:
sslmode=require built an SSL context with CERT_NONE and no hostname check, permitting an active network attacker to impersonate PostgreSQL. All encrypted modes (require, verify-ca, verify-full) now use PROTOCOL_TLS_CLIENT with certificate and hostname verification.

Upstream port of Midtown-Technology-Group/bifrost@f98b8c7de (#441).

Behavior change: self-signed DB certificates that relied on require-without-verification will now fail; that is the intended fix. Deployments with private CAs should install the CA.

Verification: tests/unit/core/test_database_tls.py 4/4 locally.


jackmusick (approved):
Authenticated TLS is now enforced consistently for encrypted database modes, with focused regression coverage for certificate and hostname verification. The intentional self-signed certificate compatibility impact is clearly documented. Looks good.

Issue #640, opened 2026-08-25
[feature]: Support multiple Bifrost instances safely in agent clients


jackmusick:
## Problem

The Bifrost skill and consolidated MCP gateway work cleanly for one connected Bifrost server, but multiple connected instances can become ambiguous. Today generated packages reuse the same plugin, skill, marketplace, and MCP identifiers while embedding one instance URL. That risks collisions and can blur production, lab, demo, organization, credential, and audit boundaries.

## Product model

Use one generic Bifrost skill with many independently named Bifrost connections. The skill may federate read-only discovery, but each MCP connection owns execution and its security boundary.

## Acceptance criteria

- [ ] Installing the generic Bifrost skill is separate from connecting a particular Bifrost instance.
- [ ] Each connection has a unique editable name plus immutable instance identity.
- [ ] Initialization exposes instance ID, display name, environment, canonical URL, and gateway contract version.
- [ ] Read-only discovery may search relevant instances and labels every result with its source instance.
- [ ] Ambiguous writes require an explicit target; writes never fan out or select the first responder.
- [ ] Agent hydration, capability/schema lookup, execution, and async polling remain pinned to one connection.
- [ ] Capability and tool references carry an `instance_ref` and are rejected on another instance even when entity UUIDs match.
- [ ] OAuth identity, permissions, required instructions, memory, and audit history remain isolated per connection.
- [ ] Projects can optionally declare a default Bifrost connection.
- [ ] Existing single-instance installations have a clear, non-destructive migration path.

## Source

Design discussion: Codex task `01a0107c-c728-75f3-bfea-822a820933b4`.

PR #773, opened 2026-09-16
fix(sdk): preserve explicit global scope

MTG-Thomas:
resolve_scope() swallowed an explicit "global" request to the default org whenever an execution context existed. Track the explicit request and resolve it to the cascade global token (None) for privileged callers, or deny with the requested value in the error for others. CLI mode (no context) still returns "global" verbatim.

Upstream port of Midtown-Technology-Group/bifrost #640/#641 (code + tests + config docs).

Verification: tests/unit/test_scope_override.py 18/18 locally.


jackmusick:
Thanks for upstreaming this. I traced the current behavior through the SDK and API, and this does not fix a missing capability on current `main`.

Today:

- Omitting `scope` uses the execution context/default organization.
- Passing `scope=None` behaves the same as omission, because the current Python signatures cannot distinguish those two calls.
- Passing `scope="global"` already works for an authorized provider/admin: the SDK preserves the literal `"global"` token, and the API translates it to the global database scope.

This change converts `"global"` to `None` before the HTTP request. At the API boundary, `None` means “scope omitted/use the caller organization,” so a global operation can silently target the organization instead.

The upstream follow-up PR #641 corrected this exact issue by preserving `scope = "global"`; that correction is missing here even though this PR says it ports both #640 and #641. With #641 applied, the combined behavioral change is effectively already what current `main` does.

If we want `scope=None` to explicitly mean global, that needs a separate sentinel-default refactor so Python can distinguish an omitted argument from an explicit `None`. This PR does not implement that distinction.

Closing because this unlocks no capability on current `main` and would regress the existing explicit-global path.

